PR #39, opened 2024-06-21
Fix crash on Vercel due to missing links in Link

heypoom:
When the `to={...}` fallback is missing, the NavLink crashes on the Vercel build, even if it runs fine on development build.

(This change is already in the release branch - syncing back to `main`)


heypoom:
superseded by https://github.com/metabase/shoppy/pull/40

PR #40, opened 2024-06-21
Design tweaks for Shoppy's new design

heypoom:
- Improves styling of login pages (mainly rounded corners and user/password input theme)
- Adds a background color above-the-fold so it loads on first load, when the stylesheet is not available
- Replace the padding workaround with CSS using new version of embedding SDK
- Fixes the CSS overrides that was causing issues with tabs in interactive dashboards
- Opens the sidebars by default (shows the product/categories and analytics side-nav)
- Fixes crash in production caused by Link component in Wouter when to={...} is undefined/null

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -13,6 +13,7 @@
     "drizzle:studio": "drizzle-kit studio"
   },
   "dependencies": {
+    "@metabase/embedding-sdk-react": "^0.1.13",
     "argon2": "^0.40.1",
     "connect-pg-simple": "^9.0.1",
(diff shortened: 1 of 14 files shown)


heypoom:
(Fixing an out-of-memory issue on Vercel before merging)

heypoom:
Created tickets for out-of-memory errors here in https://github.com/metabase/shoppy/issues/41. I'm merging this PR for now as running production builds locally works fine, so this is definitely a Vercel issue.

PR #44, opened 2024-06-27
Migrate Shoppy to Vite

### LINKED ISSUE #43: Replace create-react-app with Vite for Shoppy to improve DX
Two main reasons why this is a good idea:
- We use Shoppy as the development bench for the SDK. The startup time for Vite is almost instantaneous, compared to ~15s on create-react-app. We need to constantly restart the dev server after editing the SDK, so this would be a huge timesaver and a plus for DX.
- Vite is recommended in favor of create-react-app in the official docs. Good for external developers trying out Shoppy locally, or viewing the code for reference.

heypoom:
Closes #43 

### Changes
- Replaces create-react-app with Vite
- Uses the `yarn dev` command for both API and frontend for consistency
- Uses the `VITE_APP` prefix for environment variables instead of `REACT_APP`
- Adds SSL and Tailwind support for Vite

### Why Vite?
- Speeds up local dev server from over 15s to less than 1s. This is huge as we reload the dev server on every SDK change.
- Better reference point for SDK users, as create-react-app is considered obsolete by many.


README.md
```diff
@@ -18,7 +18,7 @@ This demo uses the data from the hosted Metabase Cloud instance and provides a h
 
 - Run the React frontend. This starts the demo application on HTTPS, which is required to connect to the hosted JWT server.
 
-  - `yarn start:https`
+  - `yarn dev:https`
```
WiNloSt, line 21: Interestingly, I've always been running `yarn start` and `yarn dev` in this PR with hosted JWT, and it still seems to work for me. Can you double-check if this is really a requirement? I mean having the HTTPS option is great in case we really need to test features that require HTTPS, but I don't want people to get the wrong idea that they only need to use this command when using hosted JWT server.

package.json
```diff
@@ -16,31 +17,33 @@
...
     "web-vitals": "^2.1.4",
     "wouter": "^3.1.2"
   },
   "scripts": {
     "install-deps": "yarn --force",
     "typecheck": "tsc -p tsconfig.json --noEmit",
-    "start": "rimraf ./node_modules/.cache && NODE_OPTIONS=--max_old_space_size=8192 PORT=3004 BROWSER=none react-scripts start",
-    "start:https": "HTTPS=true yarn start",
-    "build": "react-scripts build",
+    "dev": "NODE_OPTIONS=--max_old_space_size=8192 BROWSER=none vite",
```
WiNloSt, line 38: could you test this on your end as well? I've tried removing the old `NODE_OPTIONS` hack and it seems to work now!

postcss.config.cjs
```diff
@@ -0,0 +1,3 @@
+module.exports = {
```
WiNloSt, line 1: Do we need to have a PostCSS config now is because `react-script` use to do this automatically for us?

tailwind.config.js
```diff
@@ -1,6 +1,6 @@
 /** @type {import('tailwindcss').Config} */
 module.exports = {
-  content: ["./src/**/*.{js,jsx,ts,tsx}"],
+  content: ["./index.html", "./src/**/*.{js,jsx,ts,tsx}"],
```
WiNloSt, line 3: nit: we're likely not going to use any Tailwind classes in `index.html` do we? That file is mostly a placeholder for our main script, so I'm not sure having `index.html` here is necessary.

src/types/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
WiNloSt, line 1: From https://vitejs.dev/guide/features#client-types. It seems we either need this file or `types: ["vite/client"]` in `tsconfig.json`, so having both seems unnecessary. I've tested that after removing this file I still get types for `import.meta.env` as expected.

WiNloSt (approved):
I've left minor comments. Otherwise, this is a great change! We now can test the SDK much more conveniently.

README.md
heypoom: @WiNloSt The `dev:https` is required if you want to use our hosted API server on Vercel, without running up the API server locally (which is not possible if you are not a Metabase employee, as you need the `DB_URL`)

I decided to make that option the most prominent, as that is the only way non-employees can run the demo. In the below sections, the instruction says that Metabase employees can run locally with `yarn dev` and `cd api && yarn dev`

package.json
heypoom: nice, I tested it locally and it works without the memory increase now!

postcss.config.cjs
heypoom: @WiNloSt yes, I think react-scripts has a built-in PostCSS config, while Vite requires an explicit PostCSS configuration

src/types/vite-env.d.ts
heypoom: Oh nice, I didn't knew that the `vite/client` type is already there. Thanks!

README.md
heypoom: @WiNloSt woah, I just tried `yarn dev` without HTTPS and it now works too! I swear that never worked back when I used create-react-app, not sure what changed...

In any case, I'll update README accordingly. Thanks!

PR #45, opened 2024-06-28
Add three themes and theme switcher for Shoppy

heypoom:
Implements multiple themes support for Shoppy. Please see the demo video in the team channel.

### Screenshots

Stitch

CleanShot 2567-07-03 at 15 11 38@2x

Luminara

CleanShot 2567-07-03 at 15 11 53@2x

Pug & Play

CleanShot 2567-07-03 at 15 12 20@2x

### Sorry, this PR is huge!

Sadly, there isn't a nice way to break down this PR to make it easy to review. I've added review comments to make it easier for you. Note that most of them are CSS styles and configurations that you can glance over 🙏🏻


public/fonts/luminara.font.css
```diff
@@ -0,0 +1,28 @@
+@import url('https://fonts.googleapis.com/css2?family=Arsenal:ital,wght@0,400;0,700;1,400;1,700&family=Libre+Caslon+Display&display=swap');
+
+@font-face {
+  font-family: 'Custom';
```
heypoom, line 4: This is because how our "Custom" font family works. We can remove this once we add proper fonts support, see https://github.com/metabase/metabase/issues/44394

index.html
```diff
@@ -11,7 +11,7 @@
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
 
     <link
-      href="https://fonts.googleapis.com/css2?family=Barlow:wght@400;700&display=swap"
+      href="https://fonts.googleapis.com/css2?family=Lato:wght@400;700&display=swap"
```
heypoom, line 14: Used for the top theme switcher

api/drizzle/seeder/assign-images.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { db } from "../../src/utils/db"
+import { products as productSchema } from "../../src/schema/products"
+
+const SHOPS = ["stitch", "luminara", "pug"]
+
+/**
+ * Assign images to products based on their shops category.
+ */
+export async function assignImagesToProducts() {
```
heypoom, line 11: One-time script used to assign images. Likely will never need to be ran again, but I added it here solely for future reference.

package.json
```diff
@@ -65,6 +66,7 @@
   },
   "devDependencies": {
     "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
+    "@iconify/react": "^5.0.1",
```
heypoom, line 69: The design from Alessio uses multiple iconsets, so `iconify/react` allows us to use multiple icon sets at once.

src/components/AppProvider.tsx
```diff
@@ -26,42 +29,18 @@ const config: SDKConfig = {
...
-}
+export const AppProvider = ({ children }: Props) => {
+  const [themeKey] = useAtom($theme)
 
-export const AppProvider = ({ children }: Props) => (
-  <MetabaseProvider config={config} theme={theme}>
-    {children}
-  </MetabaseProvider>
-)
+  const theme = useMemo(() => {
```
heypoom, line 35: Metabase theme is loaded dynamically based on what theme we are selecting.

src/components/AuthCheck.tsx
```diff
@@ -1,20 +1,44 @@
...
+  })
 
-  if (query.isLoading) return <FullPageLoader />
+  const loginMutation = useMutation({
+    mutationFn: loginToSite,
+    mutationKey: ["login"],
+  })
+
+  useEffect(() => {
+    if (!query.data?.email) {
```
heypoom, line 27: If the user does not have a session yet, we login for them. This is needed for the JWT auth and SSO. We decided to simulate the login for them, so the UX is a bit easier than having to login.

src/components/RemountOnThemeChange.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useAtom } from "jotai"
+
+import { $theme } from "../store/theme"
+
+interface Props {
+  children: React.ReactNode
+}
+
+/**
+ * WORKAROUND: remount the children when the theme changes.
```
heypoom, line 11: The MetabaseProvider has no way to know that the user session has changed, so this helps to avoid a full-page reload.

src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+    return createTheme(THEME_CONFIG_MAP[themeName].mantine)
+  }, [themeName])
+
+  // apply the CSS variables to the body, depending on the theme.
+  // this ensures the variables are applied in portals as well.
+  useEffect(() => {
+    document.body.setAttribute("data-theme", themeName)
+  }, [themeName])
+
+  return <MantineProvider theme={theme}>{props.children}</MantineProvider>
```
heypoom, line 22: This is for our own app's Mantine provider - not related to the SDK.

src/components/ThemeSwitcher.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { $theme } from "../store/theme"
+
+import { THEMES } from "../themes"
+import { loginToSite } from "../utils/switch-site"
+import { ThemeKey } from "../types/theme"
+
+export const ThemeSwitcher = () => {
+  const [activeTheme, setActiveTheme] = useAtom($theme)
+
+  const loginMutation = useMutation({
```
heypoom, line 16: This handles the loading state for us. We show a loading indicator on the site change button, so the user has an immediate feedback.

src/components/ThemeSwitcher.tsx
```diff
@@ -0,0 +1,73 @@
...
+          {THEMES.map((theme) => {
+            const active = activeTheme === theme.key
+
+            return (
+              <Button
+                key={theme.key}
+                variant="outline"
+                size="xs"
+                color={active ? "#2B2F32" : "#7AC1FF"}
+                bg={active ? "#EAEAEA" : undefined}
```
heypoom, line 54: color is hard-coded here to avoid each of the 3 themes from tampering with the top-bar theme switcher (must remain the same)

src/components/layout/SidebarLinks.tsx
```diff
@@ -40,11 +39,12 @@ const renderLink = (link: SidebarLink, child?: boolean) => (
...
           cx(
-            "hover:bg-transparent",
+            "hover:bg-transparent font-sans",
             props.className,
+            !child && "sidebar-link-root",
             child && "space-y-2",
-            child && !active && "text-light-grey",
-            active ? "text-primary" : "hover:text-gray-300",
+            child && !active && "sidebar-inactive-child",
+            active ? "dark-gradient" : "sidebar-active-child",
```
heypoom, line 47: this is for themeability, as the sidebar differs a bit across themes.

src/hooks/useFontLoader.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export function FontLoader() {
+  const [theme] = useAtom($theme)
+
+  return (
+    <div>
+      <link
+        type="text/css"
+        rel="stylesheet"
+        href={`/fonts/${theme}.font.css`}
```
heypoom, line 13: we need a different set of Metabase font overrides

src/index.tsx
```diff
@@ -1,26 +1,38 @@
...
 
 import "./styles/index.css"
 
+// CSS variables used for the demo app's themes.
+import "./themes/variables.css"
+
+// Theme overrides for each themes
+import "./themes/stitch.css"
+import "./themes/luminara.css"
+import "./themes/pug.css"
```
heypoom, line 20: these are scoped with `body[data-theme]`, so it's safe to import globally

src/routes/analytics/DashboardPage.tsx
```diff
@@ -10,7 +12,13 @@ export function DashboardPage(props: Props) {
...
   return (
     <Box mih="100vh" className="dashboard-container smartscalar">
-      <InteractiveDashboard dashboardId={dashboardId} withTitle withDownloads />
+      <RemountOnThemeChange>
+        <InteractiveDashboard
+          dashboardId={dashboardId}
+          withTitle
+          withDownloads
+        />
+      </RemountOnThemeChange>
```
heypoom, line 21: Maintains the illusion of switching between themes, without forcing the entire page to go into a loading spinner.

src/routes/product-list/ProductCardFooter.tsx
```diff
@@ -0,0 +1,26 @@
+import { Box, Text, Button, Divider } from "@mantine/core"
+import { useAtom } from "jotai"
+
+import { $theme } from "../../store/theme"
+
+export function ProductCardFooter() {
+  const [theme] = useAtom($theme)
+
+  if (theme === "luminara") {
```
heypoom, line 9: Luminara uses a divider instead of button.

src/store/theme.ts
```diff
@@ -0,0 +1,6 @@
+import { atom } from "jotai"
+import { atomWithStorage } from "jotai/utils"
+
+import type { ThemeKey } from "../types/theme"
+
+export const $theme = atomWithStorage<ThemeKey>("theme", "stitch")
```
heypoom, line 6: Remembers what theme you last visited when you reload.

src/styles/index.css
```diff
@@ -6,10 +6,17 @@
...
 #metabase-sdk-root {
   background: transparent;
 }
+
+body[data-theme="luminara"] {
+  background-image: url("/light-backdrop.png");
+}
+
+body:not([data-theme="luminara"]) {
+  background: var(--color-background) !important;
```
heypoom, line 21: This is because of a bug in the SDK, specifically https://github.com/metabase/metabase/issues/44802. We delete this line once the bug has been fixed. TL;DR: EmbedFrame's style leaked.

src/styles/smartscalar-workaround.css
```diff
@@ -2,30 +2,53 @@
...
-  color: #ADABA9;
+body[data-theme="luminara"] {
+  .smartscalar [data-testid="scalar-value"] {
+    font-weight: 200;
+    font-family: "Libre Caslon Display";
+  }
+}
+
+body[data-theme="stitch"] {
+  .smartscalar [data-card-key]>div:has(div[data-testid="scalar-container"]) {
```
heypoom, line 24: Not pretty, but we have to use this workaround until the SDK officially supports fine-grained theming.

src/utils/switch-site.ts
```diff
@@ -0,0 +1,38 @@
...
+import { logout } from "./logout"
+import { queryClient } from "./query-client"
+
+const SITE_TO_USER_MAP: Record<ThemeKey, string> = {
+  stitch: "rene@example.com",
+  luminara: "cecilia@example.com",
+  pug: "emily@example.com",
+}
+
+export async function loginToSite(_site: ThemeKey | null) {
```
heypoom, line 14: Login to a new shop, so we get a sandboxed data.

src/hooks/useFontLoader.tsx
```diff
@@ -0,0 +1,17 @@
+import { useAtom } from "jotai"
+
+import { $theme } from "../store/theme"
+
+export function FontLoader() {
+  const [theme] = useAtom($theme)
+
+  return (
+    <div>
```
WiNloSt, line 9: Since it renders a `link` component, I don't think it needs to be nested under `div`. Like you could just return `link` directly.

src/hooks/useFontLoader.tsx
WiNloSt, line 1: And this file seem to be a component, but the file name was still a hook, I think we should move this file to a proper location.

src/components/AuthCheck.tsx
WiNloSt: Can you point me to where the decision has been made? I'm curious.

src/components/AuthCheck.tsx
```diff
@@ -1,20 +1,40 @@
-import { useQuery } from "@tanstack/react-query"
+import { useMutation, useQuery } from "@tanstack/react-query"
 import { Redirect } from "wouter"
 
 import { getUser } from "../utils/query-user"
 import { FullPageLoader } from "./Loader"
+import { loginToSite } from "../utils/switch-site"
+import { useEffect } from "react"
+import { ThemeKey } from "../types/theme"
+import { getCurrentTheme } from "../utils/current-theme"
```
WiNloSt, line 9: these aren't used. I'd suggest you also add ESLint to catch unused imports. It could be another PR not to make this one too big unnecessarily.

src/utils/switch-site.ts
WiNloSt, line 14: nit: I'd just rename everything relating to `theme` to `site` instead. I feels weird to use `ThemeKey` type for `site` variable, but it feels ok to use `Site` (type) for `theme` variable don't you think?

src/utils/current-theme.ts
```diff
@@ -0,0 +1,14 @@
+import { THEME_CONFIG_MAP } from "../themes"
+import { ThemeKey } from "../types/theme"
+
+const THEME_NAMES = Object.keys(THEME_CONFIG_MAP)
+
+export function getCurrentTheme(): ThemeKey {
+  const theme = JSON.parse(localStorage.getItem("theme")!)
```
WiNloSt, line 7: I find it so disconnecting that we set this value to localStorge with `atomWithStorage`

https://github.com/metabase/shoppy/blob/5c4c99b4c650f18bfaebf229004f9e7c15e85dc3/src/store/theme.ts#L6

But we read it from localStorage directly here.

I'd kind of expect you'd try to read it from the atom somehow here, otherwise, this looks like a hack to me.

I found we could use `store` to [interact with atom outside React](https://jotai.org/docs/guides/using-store-outside-react). But I'm not totally sure how to apply that to this scenario, since I'm not familiar with Jotai.

src/components/RemountOnThemeChange.tsx
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * This causes the charts data to be re-fetched and re-rendered
+ * when the theme changes.
+ */
+export function RemountOnThemeChange(props: Props) {
```
WiNloSt, line 16: This seems too overzealous in my opinion. I think we could achieve the same effect by forcing `key`.

```ts
export function RemountOnThemeChange(props: Props) {
  const [themeName] = useAtom($theme)
  return <Fragment key={themeName}>{props.children}</Fragment>
}
```

Could you see if this solves all your problems, I tried this with the dashboard page `admin/analytics/:id` and it reloaded the dashboard when changing themes

WiNloSt (commented):
This is like 80% review. I'll continue later along with any fix you might incorporate.

src/components/AuthCheck.tsx
heypoom: @WiNloSt here you go: https://metaboat.slack.com/archives/C063Q3F1HPF/p1719509315975669?thread_ts=1719295515.987739&cid=C063Q3F1HPF

WiNloSt: awesome thanks!

heypoom:
Thanks for reviewing! Converting this PR to draft for now, so I can address both Alessio's design review feedback, and your code review feedback at once 🙏🏻

src/components/AppProvider.tsx
```diff
@@ -12,6 +11,10 @@ import {
 
 import { MetabaseError, MetabaseLoader } from "./SdkStates"
 
+import { $theme } from "../store/theme"
```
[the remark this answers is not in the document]
heypoom: Good point, this was a convention from `nanostores` (a different library) rather than Jotai, let me update to `themeAtom`

src/hooks/useFontLoader.tsx
heypoom: Nice catch, I changed this file a lot so I forgot to clean it up 🙏🏻

src/hooks/useFontLoader.tsx
heypoom: Whoops, nice catch!

src/utils/current-theme.ts
heypoom: Great catch! Replaced this with `store.get` 🙏🏻

src/routes/analytics/DashboardLinkCard.tsx
```diff
@@ -12,20 +12,13 @@ export interface DashboardLinkCardProps {
...
+      <Card className="card gap-y-5 justify-between" h="100%" withBorder p={12}>
         <Box>
           <Title size="h4">{props.title}</Title>
 
-          <Text c="light-grey" fz="14px">
-            {props.description ?? "No description."}
-          </Text>
+          <Text fz="14px">{props.description ?? "No description."}</Text>
         </Box>
+        c="light-grey"
```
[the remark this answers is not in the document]
heypoom: Thanks! Looks like my IDE glitched out...

src/styles/smartscalar-workaround.css
heypoom: @deniskaber I agree with this being misleading. Let me check with Alberto what he thinks. I think Alessio's goals for the 3 themes was to [show what would be eventually possible](https://metaboat.slack.com/archives/C063Q3F1HPF/p1719336477507209?thread_ts=1719295515.987739&cid=C063Q3F1HPF) with the SDK, so this might be a necessary evil.

> these, on purpose, are pushing the possibilities of what we could do (e.g., gradients, multiple fonts)

heypoom:
NOTE: there is a pending [UI review feedback](https://www.notion.so/metabase/SDK-reference-app-Shoppy-themes-f2b539f2ea6445d8824d0caed66a6317?pvs=4#d817c03e4ad64f45a60627d30f205d60) from Alessio, but we can merge this first then address this in the next PR, so the changes wouldn't be larger than what it is.

src/store/theme.ts
WiNloSt, line 1: small thing here, but the file name is still `theme` wile all names are now `site` 🙏

src/components/SiteLogo.tsx
```diff
@@ -0,0 +1,32 @@
+import { Text, Image } from "@mantine/core"
+import { useAtom } from "jotai"
+
+import { siteAtom } from "../store/theme"
+
+export function SiteLogo() {
+  const [theme] = useAtom(siteAtom)
```
WiNloSt, line 7: nit: 
```suggestion
  const [site] = useAtom(siteAtom)
```

src/components/AppProvider.tsx
```diff
@@ -26,42 +29,18 @@ const config: SDKConfig = {
...
-    scalar: {
-      value: {
-        fontSize: "47px",
-        lineHeight: "50px",
-      },
-    },
-  },
-}
+export const AppProvider = ({ children }: Props) => {
+  const [siteKey] = useAtom(siteAtom)
```
WiNloSt, line 33: I found we name this variable from `siteAtom` 3 ways
- `siteKey`
- `siteName`
- `theme`

If you like the explicitness I'd choose either of the first two, but I think `site` is also enough.

Can we ensure we name all of these variables the same? Or do you think it it's valid to name these differently based on where the atim is used?

src/themes/index.ts
WiNloSt, line 1: nit: the folder is still named `theme`, all other theme related files could stay here, but maybe this should be moved to somewhere else like `sites/*`

src/utils/current-site.ts
WiNloSt, line 1: Great job, this isn't hacky now 🙌

WiNloSt (approved):
I only have minor comments regarding file names and variable names, otherwise, this looks great. Also tested that it's working as expected. The force reloading dashboards still works when changing sites.

awesome job 👍

src/components/AppProvider.tsx
heypoom: renamed - thanks for the keen eye!

PR #58, opened 2024-08-22
Increase interactive question height in modal

heypoom:
Fixes https://github.com/metabase/metabase/issues/46825

Increases the interactive question height in the Shoppy demo, so that it shows more content in a modal.

(Note: this issue is quite specific to Shoppy modals - our SDK components already supports `height="100%"`)

CleanShot 2567-08-22 at 20 48 00@2x


WiNloSt (commented):
Other modals look great
- image
- image

as we could see the buttons

But the notebook editor button is still buried in the modal unless we scroll
- image
- image

WiNloSt (approved):

Issue #59, opened 2024-08-27
Add ability to create and modify questions in Shoppy


heypoom:
Add the ability for users to try creating and modifying questions in Shoppy.

PR #60, opened 2024-09-02
Create questions and dashboards & update sandboxing

heypoom:
Fixes https://github.com/metabase/shoppy/issues/59

### How to Verify

Try out these flows:
- New Question > From Scratch
- New Question > From Template
- New Dashboard

Note that the "New Question" can be accessed from 2 pages: either via the buttons in the Custom page, or by the links in the sidebar navigation. Either of them will show a Menu "From Scratch" and "From Template"

### Demo

<div>
    <a href="https://www.loom.com/share/36e62352b3874be8a56d2921887c84d8">
      <p>Create Question & Create Dashboard in Shoppy - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/36e62352b3874be8a56d2921887c84d8">
      
    </a>
  </div>

### Caveats

- There are [some issues spotted on the SDK side](https://metaboat.slack.com/archives/C06FCQT0KMZ/p1725290774946069). Tasks are being created to improve these from the SDK's side.
- There are also some sandboxing-related issues that I'll tackle next, [see this thread](https://metaboat.slack.com/archives/C06FCQT0KMZ/p1725292986835669?thread_ts=1725290774.946069&cid=C06FCQT0KMZ).


heypoom:
(requesting early reviews)

src/components/layout/SidebarLinks.tsx
```diff
@@ -24,36 +15,41 @@ export function SidebarLinks() {
...
-  >
-    {link.children && link.children.map((link) => renderLink(link, true))}
-  </NavLink>
-)
+const renderLink = (link: SidebarLink, child?: boolean) => {
+  const { component: Component } = link
+
+  return (
+    <NavLink
+      label={Component ? <Component /> : link.title}
```
heypoom, line 23: Allow rendering the popover menu of "New Question"

src/constants/collections.ts
```diff
@@ -0,0 +1,15 @@
+import { SiteKey } from "../types/site"
+
+export const QUESTION_TEMPLATE_COLLECTION_ID = 42
+
+export const SANDBOXED_CUSTOM_ANALYTICS_COLLECTIONS: Record<SiteKey, number> = {
```
heypoom, line 5: These defines the 3 isolated collections for each shops.

CleanShot 2567-09-04 at 02 01 04@2x

src/constants/collections.ts
```diff
@@ -0,0 +1,15 @@
...
+  stitch: 45,
+  luminara: 46,
+  pug: 47,
+}
+
+export const SANDBOXED_USER_GENERATED_COLLECTIONS: Record<SiteKey, number> = {
```
heypoom, line 11: These defines the "User-Generated" sub-collections inside of the above collection, so we can clean-up user-generated content with a cronjob.

CleanShot 2567-09-04 at 02 01 33@2x

src/constants/sites.ts
```diff
@@ -5,9 +5,9 @@ import { PugSite } from "../themes/pug"
 import type { SiteConfig, SiteInfo, SiteKey } from "../types/site"
 
 export const SITES: SiteInfo[] = [
-  { title: "Shop 1", key: "stitch", icon: "iconoir:shirt" },
-  { title: "Shop 2", key: "luminara", icon: "streamline:lipstick" },
-  { title: "Shop 3", key: "pug", icon: "lucide:dog" },
+  { title: "Shop 1", key: "stitch", icon: "iconoir:shirt", shopId: 2 },
+  { title: "Shop 2", key: "luminara", icon: "streamline:lipstick", shopId: 3 },
+  { title: "Shop 3", key: "pug", icon: "lucide:dog", shopId: 1 },
```
heypoom, line 10: Add the `shopId` of each sites from the database.

src/routes/analytics/AnalyticsCustomPage.tsx
```diff
@@ -1,18 +1,43 @@
...
+          <Button>New Question</Button>
+        </NewQuestionMenu>
+
+        <Link
+          href="/analytics/new/dashboard"
+          onClick={() => setDashboardId(null)}
+        >
+          <Button>New Dashboard</Button>
+        </Link>
+      </Flex>
```
heypoom, line 36: Add buttons to create new question and new dashboard.

CleanShot 2567-09-04 at 02 13 56@2x

src/routes/analytics/DashboardPage.tsx
```diff
@@ -10,15 +10,12 @@ interface Props {
 export function DashboardPage(props: Props) {
   const dashboardId = parseInt(props.id, 10)
 
+  // When the site changed, reload to apply the site's sandboxed data.
+  useSiteChanged(window.location.reload)
```
heypoom, line 14: Without this, the dashboard shows sandboxed data from the old user, as there is no way for the SDK to reload the current user.

src/routes/analytics/DashboardPage.tsx
```diff
@@ -10,15 +10,12 @@ interface Props {
...
+
   return (
     <Box mih="100vh" className="dashboard-container smartscalar">
-      <RemountOnSiteChange>
```
heypoom, line 15: Using `RemountOnSiteChange` doesn't actually work here - we still get stale data from the old user.

src/routes/analytics/new/NewDashboardPage.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { siteAtom } from "../../../store/site"
+import { createDashboardIdAtom } from "../../../store/create"
+
+import { SANDBOXED_USER_GENERATED_COLLECTIONS } from "../../../constants/collections"
+
+export const NewDashboardPage = () => {
+  const [site] = useAtom(siteAtom)
+  const [dashboardId, setDashboardId] = useAtom(createDashboardIdAtom)
+
+  const collectionId = SANDBOXED_USER_GENERATED_COLLECTIONS[site]
```
heypoom, line 20: Ensures that the dashboard is saved in the "User-Generated" folder under the site's sandboxed collections by default. If the user tries to change the collection, it wouldn't work as they cannot save as per the permissions configuration.

src/routes/analytics/new/NewFromScratchPage.tsx
```diff
@@ -0,0 +1,18 @@
...
+export const NewFromScratchPage = () => {
+  const [questionKey] = useAtom(createQuestionKeyAtom)
+
+  return (
+    <Container w="100%">
+      <Title fz="28px" mb="md">
+        Create a question
+      </Title>
+
+      <CreateQuestion key={questionKey} />
```
heypoom, line 15: We should reload the CreateQuestion component whenever the "Create Question > From Scratch" button is clicked again on the side navigation.

CleanShot 2567-09-04 at 02 17 33@2x

src/routes/analytics/new/NewFromTemplatePage.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <Container>
+        <Title fz="28px" mb="md">
+          Pick a question
+        </Title>
+
+        <CollectionBrowser
+          collectionId={QUESTION_TEMPLATE_COLLECTION_ID}
+          visibleEntityTypes={["question"]}
+          onClick={(item) => setQuestionId(item.id)}
+        />
```
heypoom, line 26: Use the collection browser for selecting questions to use as templates, from the "Question Templates" collection.

CleanShot 2567-09-04 at 02 18 30@2x

src/routes/product-detail/index.tsx
```diff
@@ -1,26 +1,43 @@
...
 
+  // If the product is from a different shop, redirect back to the product listing page.
+  // This prevents the user from seeing data from a different shop.
+  useEffect(() => {
+    const site = SITES.find((s) => s.key === siteKey)
+
+    if (product && site && site.shopId !== product.shopId) {
+      navigate("/admin/products")
+    }
+  }, [siteKey, product])
```
heypoom, line 39: This is a bug in the existing Shoppy app. Go to current demo at https://metabase-shoppy.vercel.app, try to open a product and try to switch site. You'll see that we can still see products that are supposed to be in another tenant.

This makes it so if they open a URL with the product from the wrong tenant, they'll get redirected to the product listing page.

src/store/create.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const createDashboardIdAtom = atom<number | null>(null)
+export const selectedQuestionTemplateIdAtom = atom<number | null>(null)
+
+export const createQuestionKeyAtom = atom<number>(0)
+
+export const resetQuestionAtom = atom(null, (get, set) => {
+  set(createQuestionKeyAtom, get(createQuestionKeyAtom) + 1)
+  set(selectedQuestionTemplateIdAtom, null)
+})
```
heypoom, line 11: This is lifted to a global state to make it possible to reset the state from the side navigation bar.

src/themes/stitch.ts
```diff
@@ -39,7 +39,7 @@ const metabase: MetabaseTheme = {
     "text-tertiary": colors.lighterGrey,
     border: colors.darkGrey,
     background: colors.background,
-    "background-hover": "#4C4A48",
+    "background-hover": colors.background,
```
heypoom, line 42: This caused a bug in the "Recents" tab in the entity picker of the CreateQuestion component. The recents tab's background uses `background-hover`. It makes sense to use the same colors for both values for dark-themed apps.

src/types/sidebar-link.ts
```diff
@@ -1,6 +1,9 @@
 export interface SidebarLink {
   to?: string
-  title: string
+  title?: string
+  key?: string
   children?: SidebarLink[]
   defaultOpened?: boolean
+  component?: React.FC
+  onClick?: () => void
```
heypoom, line 8: Allows "New Question" and "New Dashboard" to reset its state when clicked on in the sidebar. Also allows the "From Scratch / From Template" popover menu to show on the sidebar.

src/utils/switch-site.ts
```diff
@@ -5,9 +5,9 @@ import { logout } from "./logout"
 import { queryClient } from "./query-client"
 
 const SITE_TO_USER_MAP: Record<SiteKey, string> = {
-  stitch: "rene@example.com",
-  luminara: "cecilia@example.com",
-  pug: "emily@example.com",
+  stitch: "cecilia@example.com",
+  luminara: "emily@example.com",
+  pug: "rene@example.com",
 }
```
heypoom, line 11: Fixes a bug where the product title and product category is completely mismatched in Shoppy. You'll notice in the current demos that the product titles doesn't really match the shop's title, and this is why.

PS. You will get the wrong image if you try to run the demo now. This is expected, because I put the wrong product images in the wrong shop all along 🤦🏻. I don't want to break the current Shoppy demo in production, so I will only run the script to update the product images immediately after this PR is released.

CleanShot 2567-09-04 at 02 24 30@2x

src/utils/switch-site.ts
```diff
@@ -28,8 +28,13 @@ export async function loginToSite(_site: SiteKey | null) {
     password: "password",
   })
 
+  // Wait for the session to be updated.
+  await delay(100)
```
heypoom, line 32: I was getting all sorts of errors without this delay, even though the `await login(...)` above has been awaited.

src/utils/use-site-changed.ts
```diff
@@ -0,0 +1,19 @@
+import { useAtom } from "jotai"
+import { useEffect, useRef } from "react"
+
+import { siteAtom } from "../store/site"
+
+import { SiteKey } from "../types/site"
+
+export function useSiteChanged(onSiteChanged: () => void) {
```
heypoom, line 8: Helps us to identify when the site has been changed.

heypoom (commented):
Added some notes on this PR

src/components/layout/SidebarLinks.tsx
WiNloSt: This isn't the end of the world, but we also render the divider as a link
image

src/routes/analytics/new/NewDashboardPage.tsx
WiNloSt: just a note that we don't seem to do this for when creating a question.

Somehow `Our analytics` is set by default if we don't click on the collection picker, and it will fail to save. If this needs the `CreateComponent` component to change then ignore it for this PR.

Screenshot 2024-09-06 at 4 39 54 PM

src/routes/analytics/new/NewDashboardPage.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  if (dashboardId === null) {
+    return (
+      <RemountOnSiteChange>
```
WiNloSt, line 24: based on [this comment](https://github.com/metabase/shoppy/pull/60/files?file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&show-viewed-files=true&show-deleted-files=false#r1742553220) that removes `RemountOnSiteChange`, do we still need this here?

WiNloSt (approved):
Tested. This works great. I've included minor feedback. Please act as you see fit.

src/routes/analytics/new/NewDashboardPage.tsx
heypoom: @WiNloSt Good point, I think `CreateQuestion` component doesn't have this prop as you mentioned. cc @oisincoveney do you think it make sense to have the `initialCollectionId` prop in `CreateQuestion` also, ot be consistent with `CreateDashboardModal`?

src/routes/analytics/new/NewDashboardPage.tsx
heypoom: @WiNloSt yes, this is just to reload the "Which collection should this go in?" dropdown. Otherwise, the collection dropdown still shows the old set of collections.

src/components/layout/SidebarLinks.tsx
heypoom: @WiNloSt fixed in https://github.com/metabase/shoppy/pull/60/commits/ad523b0ba5c105d688155ab4888ea86d5646705f, thanks!

Issue #62, opened 2024-10-02
Improve mobile responsiveness for Shoppy demo


heypoom:
Make the Shoppy app responsive on mobile.

PR #65, opened 2024-10-03
Fix basic mobile responsiveness issues (Part 1)

heypoom:
Fixes https://github.com/metabase/shoppy/issues/62

Improves very basic usability on mobile, such as the navigation bar, site switcher and columns. This is only the first pass and so only fixes basic glaring responsiveness issues - and does not consider the design itself, I'll pair with Alessio to improve the design aesthetics and the rest.

# Demo

### DIFF
--- a/src/components/layout/SidebarLinks.tsx
+++ b/src/components/layout/SidebarLinks.tsx
@@ -21,10 +21,10 @@ export function SidebarLinks({ onLinkClick }: SidebarLinkProps) {
 
 const renderLink = (
   link: SidebarLink,
-  context: { child?: boolean; onLinkClick?: () => void },
+  context: { isChild?: boolean; onLinkClick?: () => void },
 ) => {
   const { component: Component } = link
-  const { child, onLinkClick } = context ?? {}
+  const { isChild, onLinkClick } = context ?? {}
 
   if (link.isDivider) {
     return (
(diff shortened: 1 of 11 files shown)


src/components/layout/SidebarLinks.tsx
```diff
@@ -60,7 +73,10 @@ const renderLink = (link: SidebarLink, child?: boolean) => {
       )}
       defaultOpened={link.defaultOpened}
     >
-      {link.children && link.children.map((link) => renderLink(link, true))}
+      {link.children &&
+        link.children.map((link) =>
+          renderLink(link, { child: true, onLinkClick }),
```
WiNloSt, line 78: nit: I'd rename this to sound more like a boolean, `child` or `children` could be confused with a React node.

```suggestion
          renderLink(link, { isChild: true, onLinkClick }),
```

WiNloSt (approved):
Looks good to me 👍

src/components/layout/SidebarLinks.tsx
heypoom: Nice catch, fixed in aa9757619c6cdc04af1f7102e4437be59b9f5afd. Thanks!

heypoom:
Note: DO NOT MERGE THIS PR YET. There is a bug with the site switching logic, causing infinite page refreshes. Will fix that first.

heypoom:
PS. we'll merge this as-is first, and work on design improvements in the follow-up PR per [this Notion document](https://www.notion.so/metabase/Mobile-Design-Pairing-with-Alessio-11569354c90180559b6be88aaee6539b?pvs=4).

heypoom:
Merging this first -- I'll address further review comments (if any) in the next PR 👍🏻

PR #66, opened 2024-10-03
Allow metabase.com and vercel.app to access Shoppy APIs

heypoom:
Allow metabase.com and vercel.app to access Shoppy APIs for the hosted demos and deploy previews.

This is to allow https://embedded-analytics-sdk-demo.metabase.com/ and Vercel deploy previews to access the API on the mock server.

### DIFF
--- a/api/src/middleware/index.ts
+++ b/api/src/middleware/index.ts
@@ -7,8 +7,7 @@ import { FRONTEND_URL, SESSION_SECRET, VERCEL_ENV } from "../constants/env"
 import { pg } from "../utils/db"
 
 // Allow these origins to access the mock API server.
-const isWhitelistedOrigin = (origin?: string) =>
-  !origin ||
+const isWhitelistedOrigin = (origin: string) =>
   origin.includes("localhost") ||
   origin.includes("vercel.app") ||
   origin.includes("metabase.com") ||
@@ -21,11 +20,8 @@ export function setupMiddleware(app: Express) {
 
   const corsMiddleware = cors({
     origin: (origin, callback) => {
-      if (isWhitelistedOrigin(origin)) {
-        return callback(null, true)
-      }
-
-      callback(new Error("cross-origin request not allowed"))
+      // Return the CORS header if the origin is defined in header and whitelisted.
+      callback(null, origin && isWhitelistedOrigin(origin))
     },
     credentials: true,
   })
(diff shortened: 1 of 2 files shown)


api/src/middleware/index.ts
```diff
@@ -6,26 +6,22 @@ import pgSessionStore from "connect-pg-simple"
 import { FRONTEND_URL, SESSION_SECRET, VERCEL_ENV } from "../constants/env"
 import { pg } from "../utils/db"
 
+// Allow these origins to access the mock API server.
+const isWhitelistedOrigin = (origin?: string) =>
+  !origin ||
```
WiNloSt, line 11: If `origin` is falsy then this function would return true. I'm not sure if that's the intention, but it feels instinctively wrong.

WiNloSt (commented):
And can you give some more context why we need this PR?

api/src/middleware/index.ts
npretto: I think origin is only falsy when the request is not coming from a browser.
That said, aren't all shoppy apis GET only? We could just allow any origin in that case

heypoom: @WiNloSt When the request is not coming from a browser, e.g. when you `curl` there is no `Origin` header therefore it should allow any origin by default. This is the default behaviour for all CORS implementation. This is just to allow https://embedded-analytics-sdk-demo.metabase.com and Vercel deploy previews to access the API.

We also have the `/metabase/sso` APIs which is not GET only.

WiNloSt: Well if you use curl you don't need CORS header anyway doesn't it? CORS headers are only needed when the request is originated from browsers.

WiNloSt (approved):

api/src/middleware/index.ts
heypoom: @WiNloSt That's a very good point. Let me see if we can somehow omit the CORS header from the dynamic CORS implementation without throwing an error using the `cors` module from Express. I wonder if using `callback(null, false)` will do it...

WiNloSt: Well maybe slapping a comment here could be enough. I don't want it to sound too complicated since this is a very low-traffic file.

heypoom: @WiNloSt fixed in https://github.com/metabase/shoppy/pull/66/commits/9d8c5f2c27a2d2370b02d0726f94270916242b44, I just realized we can simply use `callback(null, origin && isWhitelistedOrigin(origin))` and it work send the CORS origin only if it is defined and whitelisted, and not send the header otherwise.

CleanShot 2567-10-04 at 22 15 42@2x

PR #68, opened 2024-10-04
Improve design on mobile and tablet for Shoppy

heypoom:
See [this Notion doc](https://www.notion.so/metabase/Mobile-Design-Pairing-with-Alessio-11569354c90180559b6be88aaee6539b?pvs=4) for more information.

# Changes

- [x]  remove max-w-[320px] on Shell.tsx
- [x]  header font size: 36 on mobile, same for all shops
- [x]  mobile: full width product image, no padding (mobile only)
- [x]  site switcher: full width on mobile (mobile only)
- [x]  sidenav mobile: reduce padding to be the same as mobile
- [x]  pug & play mobile: all blue background, no wavey lines
- [x]  hide create question and dashboard on mobile
- [x]  reduce padding for analytics overview, all sites
- [x]  shop 3 > dashboard pages — put a rounded border (similar to Analytics Overview card)
    - [x]  16px px padding - base+
    - [x]  24px px padding - xs+
- [x]  remove orders and campaigns link — we don’t have this
- [x]  bug
    - [x]  z-index of the "new dashboard" modal is less than the side navigation (shop 3), shows up on Tablet view
    - [x]  New Question on mobile should not close the sidenav on mobile


src/components/SiteSwitcher.tsx
```diff
@@ -26,16 +26,22 @@ export const SiteSwitcher = () => {
...
       </Text>
 
-      <ButtonGroup variant="outline">
+      <ButtonGroup variant="outline" w={{ base: "100%", sm: "auto" }}>
         {SITES.map((site) => {
           const active = currentSite === site.key
 
           return (
             <Button
+              w={{ base: "100%", sm: "auto" }}
```
WiNloSt, line 44: It seems the combination of `w` on the `ButtonGroup` and here accidentally got rid of the left button rounded border.

Screenshot 2024-10-07 at 2 08 47 PM

src/components/layout/Shell.tsx
```diff
@@ -29,11 +29,14 @@ export function Shell(props: Props) {
           navbar: "navbar overflow-scroll sm:overflow-visible",
         }}
       >
-        <AppShell.Header className="border-b-[#55595B]" zIndex={102}>
+        <AppShell.Header
+          zIndex={102}
+          bg="#2B2F32"
+          className="border-b-[#2B2F32]"
```
WiNloSt, line 35: I couldn't find this change noted [in the doc](https://www.notion.so/metabase/Mobile-Design-Pairing-with-Alessio-11569354c90180559b6be88aaee6539b), but it makes sense that the border is the same color as the background. But maybe `border: none` or `border-color: transparent`, which are simpler, would do the trick with little to none visual difference.

src/components/layout/Shell.tsx
```diff
@@ -42,39 +45,38 @@ export function Shell(props: Props) {
           >
             <Image src="/metabase-logo-with-wordmark.svg" />
 
-            <Box className="md:hidden">
+            <Flex display={{ base: "flex", sm: "none" }}>
```
WiNloSt, line 48: This was quite confusing when changing from `md` to `sm`, because we changed from tailwind breakpoint to Mantine breakpoint, but they have the same value.

I wonder if we really need to wrap the burger menu inside a flex since it's the only child. And seems like we can pass `display` prop to `Burger` directly since it's also a Mantine component.

I tried with this and it seems to work (without `Flex`). This simplifies markup a bit.

```ts
<Burger
  display={{ sm: "none" }}
  opened={isMobileNavOpen}
  onClick={toggleMobileNav}
  aria-label="Toggle navigation"
  color="#eee"
/>
```

src/components/layout/Shell.tsx
```diff
@@ -42,39 +45,38 @@ export function Shell(props: Props) {
...
                 opened={isMobileNavOpen}
                 onClick={toggleMobileNav}
                 aria-label="Toggle navigation"
                 color="#eee"
               />
-            </Box>
+            </Flex>
 
-            <Box className="hidden md:block">
+            <Flex display={{ base: "none", sm: "flex" }}>
```
WiNloSt, line 57: What's the point of using `Flex` instead of `Box` if you're only specifying `display` and controlling the `flex` value manually? Is there anything not-obvious I missed here.

src/components/layout/Shell.tsx
```diff
@@ -42,39 +45,38 @@ export function Shell(props: Props) {
...
             </Flex>
           )}
         </AppShell.Header>
 
-        <AppShell.Navbar
-          p="md"
-          withBorder={false}
-          pt={107}
-          pl="30px"
-          zIndex={3}
```
WiNloSt, line 71: I'm not sure if this is addressing the z-index issue, but I still see it on a tablet view.
image

src/components/layout/SidebarLinks.tsx
```diff
@@ -33,6 +33,7 @@ const renderLink = (
         w="140px"
         my="14px"
         key={link.key}
+        className="hide-on-mobile"
```
WiNloSt, line 36: why do you use `.hide-on-mobile` and use Mantine breakpoint with `display` to do that instead in `src/components/layout/Shell.tsx`? It seems you're solving the same problem in multiple ways. Can we use only one instead of two?

src/themes/luminara.css
```diff
@@ -72,7 +72,10 @@ body[data-theme="luminara"] {
 
   .product-detail-image {
     border-radius: 16px;
-    max-width: 300px;
+
+    @media screen and (min-width: 576px) {
```
WiNloSt, line 76: Not sure why we want this on only 2 themes as it's mentioned the image should be full width.
image

Full width
image

src/themes/pug.css
```diff
@@ -13,7 +13,10 @@ body[data-theme="pug"] {
 
   .product-detail-image {
     border-radius: 100%;
-    max-width: 300px;
+    
+    @media screen and (min-width: 576px) {
```
WiNloSt, line 17: This causes the same as mentioned above. I think we should remove this so the image could be full-width.

src/components/layout/Shell.tsx
heypoom: The links to "new dashboard" is not supposed to show up on tablet view, strange. z-index is quite tricky...

src/themes/luminara.css
heypoom: @WiNloSt The idea from discussing with Alessio is that we should only make it full-width on the mobile (base) viewport, therefore `xs` and above should start to constrain the image's width. Otherwise it looks way too big on screens larger than xs (576px).

So, the fix should actually be to also make the `product-detail-image` CSS global (as opposed to being specific to theme), so it also applies to theStitch (the dark theme)

src/themes/pug.css
heypoom: @WiNloSt see https://github.com/metabase/shoppy/pull/68#discussion_r1790181614, the idea is that the image is only full width below 576px, and cap the width after that.

src/components/layout/Shell.tsx
heypoom: @WiNloSt We could've also used `Box` here, this is just to use `Flex` everywhere in the Shell component. I could revert this back to Box.

src/components/layout/Shell.tsx
heypoom: @WiNloSt Tailwind breakpoint is actually slightly different than Mantine's breakpoint, so there are subtle things that break on specific viewport sizes if we use Tailwind's breakpoints. Let me try out your markup instead, thanks!

src/themes/luminara.css
WiNloSt: OK, so I misread that style, still the issue is that themes other than the first site don't have full-width images on product detail page which should be fixed.

src/components/layout/Shell.tsx
heypoom: @WiNloSt Let me try with `border-none`, thanks! This change was because of a visual artifact of adding the site switcher - there is a tiny grey border separating the navbar and the site switcher on mobile viewports.

src/themes/luminara.css
heypoom: @WiNloSt not sure what viewport sizes you are using, but the design requirement is to use full-width images **only on viewport below 576px**, which it currently does.

Viewport of 544px:

CleanShot 2567-10-07 at 20 07 35@2x

Viewport of 642px:

CleanShot 2567-10-07 at 20 07 39@2x

The above is expected. So, it's correct on Luminar and Pug & Play, but not on theStitch (dark theme)

WiNloSt: > @WiNloSt not sure what viewport sizes you are using, but the design requirement is to use full-width images **only on viewport below 576px**, which it currently does.
> 
> Viewport of 559px:
> 
> CleanShot 2567-10-07 at 20 07 35@2x
> 
> Viewport of 681px:
> 
> CleanShot 2567-10-07 at 20 07 39@2x
> 
> The above is expected. So, it's correct on Luminar and Pug & Play, but not on theStitch (dark theme)

That explains it. Thanks! I think the viewport I use might be larger than that, so the I mistook the full-width on Stitch theme as the correct one, when it's the one that needs to be fixed.

src/components/layout/Shell.tsx
heypoom: Fixed in https://github.com/metabase/shoppy/pull/68/commits/afe401bba8034afa946b57054e6aab40fb2fd494

src/components/SiteSwitcher.tsx
heypoom: Fixed in https://github.com/metabase/shoppy/pull/68/commits/b10750cc499f1836f48a2187c687a3b52743a814 by using CSS, to avoid Mantine prepending the `<style>` tag to the first button

src/components/layout/Shell.tsx
heypoom: Fixed in https://github.com/metabase/shoppy/pull/68/commits/a60d57c11fc38ada2df06eb7a368e382cff0e408

src/components/layout/Shell.tsx
heypoom: Fixed in https://github.com/metabase/shoppy/pull/68/commits/04e252be32882e22d79e2f33e7fc861a8819b80d

src/themes/pug.css
heypoom: Fixed in https://github.com/metabase/shoppy/pull/68/commits/c34eb0cec20ce88c3c4fde2cbe0d09f62499ef95

src/themes/luminara.css
heypoom: Fixed in https://github.com/metabase/shoppy/pull/68/commits/c34eb0cec20ce88c3c4fde2cbe0d09f62499ef95

src/components/layout/SidebarLinks.tsx
heypoom: Fixed in https://github.com/metabase/shoppy/pull/68/commits/1476534d34545c4a1c3076663a715200ad25e5aa to use `hide-on-mobile` consistently.

src/components/layout/Shell.tsx
heypoom: Fixed in https://github.com/metabase/shoppy/pull/68/commits/faf78d2757f972e0914d324a17d1758e9ec119de - the zIndex was removed because I thought it was only needed for the "See more" button in theStitch (which I fixed in this PR), but turns out I still need it for the New Dashboard modal to be on top of the navbar.

CleanShot 2567-10-08 at 03 02 14@2x

heypoom:
Thanks for reviewing @WiNloSt! I addressed all of the review feedback above 🙏🏻

WiNloSt (approved):
Looks good to me 👍

Issue #70, opened 2024-10-07
Fix lighthouse performance reports on Shoppy


heypoom:
Review the Lighthouse performance reports in Shoppy, and fix them.

PR #71, opened 2024-10-08
Cache data endpoints, optimize static assets & reserve image size

heypoom:
Fixes https://github.com/metabase/shoppy/issues/70

- Caches the rarely-changing data endpoints (i.e. `/api/products` and `/api/categories`) with `Cache-Control` headers and passing a query parameter, both on the browser and on [Vercel's Edge Network](https://vercel.com/docs/edge-network/caching)
- Get rid of cookies, and replace them with using query parameters for cacheability
- Preload static assets with `<link rel="preload">`
- Prevent layout shifts by reserving the image size using aspect-ratio, object-fit and max-width
- Avoid render-blocking font loading by using Vite webfont plugin
- Optimize images and SVGs by using the Image Optimizer plugin

### Demos

Image Optimizer result

CleanShot 2567-10-08 at 09 37 25@2x


api/src/routes/category-list.ts
```diff
@@ -22,6 +22,9 @@ export async function categoryListHandler(req: Request, res: Response) {
       where: (category, { eq }) => eq(category.shopId, user.shopId),
     })
 
+    // cache for 1 week
+    res.set("Cache-Control", "public, max-age=604800")
```
heypoom, line 26: These data endpoints almost never changes, so it's a great candidate for caching. The frontend calls this endpoint with a `?site=...` query parameter, so the cache is isolated by site.

index.html
```diff
@@ -10,9 +10,47 @@
     <link rel="preconnect" href="https://fonts.googleapis.com" />
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
 
+    <!-- preload product images before the bundle loads -->
```
heypoom, line 13: A bit ugly, but it's the most surefire way of ensuring that we preloaded our product images before the JavaScript bundle even loads.

index.html
```diff
@@ -10,9 +10,47 @@
...
+    <link rel="preload" as="image" type="image/png" href="/images/pug/3.png" />
+
+    <!-- preload the Lato font -->
     <link
       href="https://fonts.googleapis.com/css2?family=Lato:wght@400;700&display=swap"
-      rel="stylesheet"
+      rel="preload"
+      as="style"
+      onload="this.onload=null;this.rel='stylesheet'"
     />
```
heypoom, line 54: This is a trick that preloads the CSS stylesheet for the Lato font first, then turns it into a stylesheet once it finishes loading. Makes the font loads a bit faster.

src/components/layout/Shell.tsx
```diff
@@ -39,7 +39,9 @@ export function Shell(props: Props) {
             className="border-transparent"
             ff="Lato"
           >
-            <Image src="/metabase-logo-with-wordmark.svg" />
+            <Link to="/">
+              <Image src="/metabase-logo-with-wordmark.svg" />
+            </Link>
```
heypoom, line 44: Take the user back to their homepage when clicking on the Metabase logo.

src/index.tsx
```diff
@@ -26,10 +26,10 @@ import "./styles/viz-workaround.css"
 
 import { store } from "./store"
 
-import { setInitialSiteCookie } from "./utils/switch-site"
+import { syncInitialSiteCookie } from "./utils/switch-site"
 import { ThemeProvider } from "./components/ThemeProvider"
 
-setInitialSiteCookie()
+syncInitialSiteCookie()
```
heypoom, line 32: This is from https://github.com/metabase/shoppy/pull/69

src/routes/product-list/index.tsx
```diff
@@ -19,8 +19,8 @@ export const ProductAnalyticsPage = (props: Props) => {
   const categoryId = props.categoryId && parseInt(props.categoryId, 10)
 
   const query = useQuery({
-    queryKey: ["products"],
-    queryFn: () => getProductList(),
+    queryKey: ["products", site],
+    queryFn: () => getProductList(site),
```
heypoom, line 23: Cache the product by site key.

src/utils/query-category.ts
```diff
@@ -1,9 +1,10 @@
 import type { Category } from "../types/category"
 
 import { API_HOST } from "../constants/env"
+import { SiteKey } from "../types/site"
 
-export async function getCategoryList(): Promise<Category[]> {
-  const response = await fetch(`${API_HOST}/categories`, {
+export async function getCategoryList(site: SiteKey): Promise<Category[]> {
+  const response = await fetch(`${API_HOST}/categories?site=${site}`, {
```
heypoom, line 7: `?site` here acts as a cache key.

src/utils/switch-site.ts
```diff
@@ -18,10 +18,16 @@ const SITE_TO_EMAIL_MAP: Record<SiteKey, string> = {
...
+export const getCurrentSite = (): SiteKey =>
+  JSON.parse(localStorage.getItem(SITE_KEY)!) || DEFAULT_SITE
+
+/**
+ * Sync the user cookie for the API with the site in localStorage.
+ *
+ * Uses the site from localStorage as the initial user cookie.
+ * This is a migration for people who have previously visited the demo.
+ **/
+export const syncInitialSiteCookie = () => setSiteCookie(getCurrentSite())
```
heypoom, line 30: This is from https://github.com/metabase/shoppy/pull/69

api/src/utils/cache.ts
```diff
@@ -0,0 +1,8 @@
+import { Response } from "express"
+
+// cache for 1 week
+export const cacheStaticResponse = (res: Response) =>
+  res.set(
+    "Cache-Control",
+    "public, max-age=604800, s-maxage=604800, stale-while-revalidate=18000",
+  )
```
heypoom, line 8: This cache-control header tells the browser to cache it for a week, as well as tells Vercel's Edge CDN to cache it as per https://vercel.com/docs/edge-network/caching.

api/src/utils/cache.ts
```diff
@@ -0,0 +1,10 @@
+import { Response } from "express"
+
+// cache for 1 week.
+// this cache-control header tells the browser to cache it for a week,
+// as well as tells Vercel's Edge CDN to cache via https://vercel.com/docs/edge-network/caching
+export const cacheStaticResponse = (res: Response) =>
+  res.set(
+    "Cache-Control",
+    "public, max-age=604800, s-maxage=604800, stale-while-revalidate=18000",
+  )
```
heypoom, line 10: The frontend uses the `?site=...` query parameter in the URL, therefore it is safe to cache this in Vercel's CDN.

public/shoppy/luminara/1.png
heypoom, line 1: The images under `public/shoppy` is unused. The correct on is `public/images/`. We remove this so that the Vite Image Optimizer does not waste time on optimizing these.

src/utils/query-category.ts
```diff
@@ -1,9 +1,10 @@
 import type { Category } from "../types/category"
 
 import { API_HOST } from "../constants/env"
+import { SiteKey } from "../types/site"
 
-export async function getCategoryList(): Promise<Category[]> {
-  const response = await fetch(`${API_HOST}/categories`, {
+export async function getCategoryList(site: SiteKey): Promise<Category[]> {
```
npretto, line 6: I'm starting to wonder if at this point we could just make the requests a static return on the FE as this is a "fake" app

heypoom: @npretto I've considered that too, but the same information is also needed for the embedded questions, e.g. we show the products and categories in the dashboard. Having to duplicate the data in both the database and the frontend is a bit iffy, as we have many products.

npretto: Ah I didn't realize we queried the same db from metabase, all good then, the cache should be enough :)

src/index.tsx
npretto: Is it needed on this pr?

api/src/utils/cache.ts
```diff
@@ -0,0 +1,10 @@
+import { Response } from "express"
+
+// cache for 1 week.
+// this cache-control header tells the browser to cache it for a week,
+// as well as tells Vercel's Edge CDN to cache via https://vercel.com/docs/edge-network/caching
+export const cacheStaticResponse = (res: Response) =>
```
npretto, line 6: The name is a bit confusing as I was thinking this function stored the request but instead it's just setting the headers for cache control.
Thinking a bit more about it though maybe we can set the headers on a middleware? I think we want to cache everything outside of the auth right?

api/src/routes/product-list.ts
```diff
@@ -24,6 +25,8 @@ export async function productListHandler(req: Request, res: Response) {
       where: (products, { eq }) => eq(products.shopId, user.shopId),
     })
 
+    cacheStaticResponse(res)
```
npretto, line 28: This endpoint is behind the `restrict` middleware for auth, does the cache mess with this?
Bonus question: do we really need the restrict?

src/index.tsx
heypoom: @npretto removed from this PR in https://github.com/metabase/shoppy/pull/71/commits/4e688df9448b307b4b8c1f3a495026e3d9bf4ed6 - as we use query parameters for cacheability instead of cookies, I got rid of cookies entirely.

src/utils/switch-site.ts
heypoom: Removed as of https://github.com/metabase/shoppy/pull/71/commits/4e688df9448b307b4b8c1f3a495026e3d9bf4ed6 as we use query parameters instead of cookies for cacheability of `/products` and `/categories` endpoint

api/src/routes/product-list.ts
heypoom: @npretto Good point. I removed the `restrict` middleware, and removed usage of cookies and rely only on query parameters. That should make it much safer to cache. Fixed in https://github.com/metabase/shoppy/pull/71/commits/4e688df9448b307b4b8c1f3a495026e3d9bf4ed6

api/package.json
```diff
@@ -13,7 +13,6 @@
     "drizzle:studio": "drizzle-kit studio"
   },
   "dependencies": {
-    "cookie-parser": "^1.4.6",
```
heypoom, line 16: Removed our reliance on cookie, as using query parameters makes it cacheable with [Vercel's Edge Network](https://vercel.com/docs/edge-network/caching).

api/package.json
```diff
@@ -25,8 +24,8 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.4.1",
-    "@types/cookie-parser": "^1.4.7",
     "@types/cors": "^2.8.17",
+    "@types/express": "^4",
```
heypoom, line 28: Fixes type issues in `main.ts`.

api/src/constants/users.ts
```diff
@@ -1,31 +1,25 @@
 import { User } from "../types/user"
 
-/** Hash of the "password" password, used for the demo. */
-const DEMO_PASSWORD_HASH = `$argon2id$v=19$m=65536,t=3,p=4$SPxNs6dPuj0kL0HV/Q+EaQ$sxUyyap4NFEDweTLJZaoLbuVGZGTD297Dz+Hh3Jvahs`
```
heypoom, line 4: We don't need this anymore as we don't have login and logout anymore.

api/src/main.ts
```diff
@@ -16,10 +16,10 @@ setupMiddleware(app)
...
 const router = Router()
 router.get("/", (_, res) => res.send({ status: "ok" }))
-router.get("/sso/metabase", restrict, metabaseAuthHandler)
-router.get("/products", restrict, productListHandler)
-router.get("/categories", restrict, categoryListHandler)
-router.get("/product/:id", productDetailHandler)
+router.get("/sso/metabase", metabaseAuthHandler)
+router.get("/products", withCacheHeader, productListHandler)
+router.get("/categories", withCacheHeader, categoryListHandler)
+router.get("/product/:id", withCacheHeader, productDetailHandler)
```
heypoom, line 22: Use the `withCacheHeader` middleware on cacheable data endpoints. These endpoints are all safe to cache now, as we got rid of cookies.

api/src/middleware/cache.ts
```diff
@@ -0,0 +1,18 @@
...
+ * and Vercel's Edge CDN caches it via https://vercel.com/docs/edge-network/caching
+ */
+export function withCacheHeader(
+  _req: Request,
+  res: Response,
+  next: NextFunction,
+) {
+  res.set(
+    "Cache-Control",
+    "public, max-age=604800, s-maxage=604800, stale-while-revalidate=18000",
```
heypoom, line 14: `s-maxage` and `stale-while-revalidate` is for Vercel, while `max-age` is for the Browser

api/src/middleware/index.ts
```diff
@@ -25,8 +24,6 @@ export function setupMiddleware(app: Express) {
   app.use(express.json())
   app.use(express.text())
   app.use(express.urlencoded({ extended: false }))
-  app.use(cookieParser())
```
heypoom, line 28: Removed the cookie parser as we do not need to parse cookies anymore.

api/src/routes/category-list.ts
```diff
@@ -19,7 +24,7 @@ export async function categoryListHandler(req: Request, res: Response) {
         description: true,
         createdAt: true,
       },
-      where: (category, { eq }) => eq(category.shopId, user.shopId),
+      where: (category, { eq }) => eq(category.shopId, shopId),
```
heypoom, line 27: We only need the shop id for querying, so we can use `getShopIdBySite` instead to simplify things.

api/src/routes/metabase-sso.ts
```diff
@@ -7,11 +7,16 @@ import {
...
 
 export async function metabaseAuthHandler(req: Request, res: Response) {
-  const user = await findUserByEmail(req.cookies.user)
+  const { site } = req.query
+
+  if (typeof site !== "string") {
+    return res.status(400).json({ error: "site is not specified" })
+  }
+
+  const user = await getUserBySite(site)
```
heypoom, line 19: We use the `site` query parameter to know which user to sign the JWT token with. This endpoint is **not** cached.

api/src/routes/product-list.ts
```diff
@@ -21,7 +26,7 @@ export async function productListHandler(req: Request, res: Response) {
         shopId: true,
       },
       with: { category: { columns: { id: true, name: true } } },
-      where: (products, { eq }) => eq(products.shopId, user.shopId),
+      where: (products, { eq }) => eq(products.shopId, shopId),
```
heypoom, line 29: Same as above. We only need the shop id for querying, so we can use `getShopIdBySite` instead to simplify things.

api/src/utils/sites.ts
```diff
@@ -0,0 +1,20 @@
...
+export const SITE_TO_SHOP_IDS: Record<string, number> = {
+  pug: 1,
+  stitch: 2,
+  luminara: 3,
+}
+
+export const getShopIdBySite = (site: string): number | undefined =>
+  SITE_TO_SHOP_IDS[site]
+
+export const getUserBySite = (site: string) => {
```
heypoom, line 12: This is needed for the `/metabase/sso` endpoint, so we know which user to use for signing the JWT, to send to the Metabase instance.

src/components/AppProvider.tsx
```diff
@@ -19,23 +19,27 @@ interface Props {
...
+  const config: SDKConfig = useMemo(() => {
+    return {
+      metabaseInstanceUrl: METABASE_INSTANCE_URL,
+      jwtProviderUri: `${API_HOST}${JWT_PROVIDER_URI}`,
+      loaderComponent: MetabaseLoader,
+      errorComponent: MetabaseError,
+
+      // Append the current site as a query parameter to the JWT provider URL.
+      fetchRequestToken: (url: string) =>
+        fetch(`${url}?site=${site}`).then((res) => res.json()),
```
heypoom, line 39: We need the `site` query parameter in `/metabase/sso` to change depending on which site we are on, so we can wrap the SDK configuration in a `useMemo`.

src/components/SiteSwitcher.tsx
```diff
@@ -14,14 +14,15 @@ import { SiteKey } from "../types/site"
 export const SiteSwitcher = () => {
   const [currentSite, setCurrentSite] = useAtom(siteAtom)
 
-  const switchSiteMutation = useMutation({
-    mutationFn: switchSite,
-    mutationKey: ["login"],
+  const refetchSiteContentMutation = useMutation({
```
heypoom: Rename to reflect what it should do: re-fetch the products and categories whenever the site changes.

src/routes/product-list/ProductCard.tsx
```diff
@@ -26,7 +26,17 @@ export const ProductCard = ({ product }: Props) => {
...
+            w="100%"
+            className="product-card-image-container aspect-ratio-square"
+          >
+            <Image
+              src={image}
+              className="product-card-image object-cover"
+              w="100%"
+              h="100%"
+            />
+          </Box>
```
heypoom, line 39: Fixes layout shifts (CLS) by reserving width and height for the product images.

src/themes/pug.css
```diff
@@ -7,7 +7,6 @@ body[data-theme="pug"] {
 
   .product-card-image {
     border-radius: 100%;
-    max-width: 300px;
```
heypoom, line 10: This should apply to every site.

src/utils/query-category.ts
heypoom, line 7: We use `?site` query parameter here to specify which site we are on, and to provide a stable identifier for caching.

vite.config.ts
```diff
@@ -2,11 +2,19 @@ import { defineConfig } from "vite"
...
 const isHTTPS = process.env.HTTPS
 
 export default defineConfig({
-  plugins: [react(), viteTsconfigPaths(), isHTTPS ? basicSsl() : null],
+  plugins: [
+    react(),
+    viteTsconfigPaths(),
+    isHTTPS ? basicSsl() : null,
+    webfontDownload(),
+    ViteImageOptimizer(),
```
heypoom, line 16: Makes webfont non-blocking, and optimizes SVG and JPEG/PNG images.

src/components/SiteSwitcher.tsx
```diff
@@ -2,28 +2,14 @@ import { Flex, Text, ButtonGroup, Button } from "@mantine/core"
...
 
-  const switchSiteMutation = useMutation({
-    mutationFn: switchSite,
-    mutationKey: ["login"],
-  })
-
-  async function changeSite(key: SiteKey) {
-    await switchSiteMutation.mutateAsync(key)
-    setCurrentSite(key)
-  }
```
heypoom, line 25: This is automatically handled by `@tanstack/react-query` when we provided the query key, e.g. `queryKey: ["categories", site]`

src/routes/product-list/ProductCard.tsx
WiNloSt: How do I verify this works? I tried removing `src` from the image and it seems the container shrinks.

image

Also, we might not need this change since we already preload all product images.

WiNloSt (approved):
Overall this looks good. I've commented inline.

src/components/AppProvider.tsx
npretto: From the point of view of the manteiners of the sdk, i think it's safer/easier to keep `jwtProviderUri` constant and ask for all the dynamic logic in `fetchRequestToken`

WiNloSt: > From the point of view of the manteiners of the sdk, i think it's safer/easier to keep `jwtProviderUri` constant and ask for all the dynamic logic in `fetchRequestToken`

This isn't well communicated in our Readme though. If I'm the customer I'd expect an easy URL manipulation via `jwtProviderUri`.

Would it be hard to make that happen?

npretto: It's not communicated at all and it probably even works, but I'm not sure we should tell people to do that, as I think it's easier for us to tell them that if they need anything dynamic they should use the function so that we only have to test that to work.
It's just for the sake of simplicity, if you feel it's more natural to have a dynamic url i'm okay with that but we'll have to write tests for both and make sure both ways work and will keep working in the future

WiNloSt: I think this could be a side-discussion. This doesn't have to be addressed here.

src/routes/product-list/index.tsx
```diff
@@ -21,20 +21,20 @@ export const ProductAnalyticsPage = (props: Props) => {
...
-
   let products = query.data ?? []
 
   if (categoryId) {
     products = products.filter((product) => product.category.id === categoryId)
   }
 
+  // If the site changes, redirect back to the product listing page.
+  // This ensures we don't show product from last site's categories.
+  useSiteChanged(() => navigate("/admin/products"))
```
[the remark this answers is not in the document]
heypoom: @WiNloSt It's just a merge conflict. I added this block in this PR but also in the mobile responsiveness PR, so once we merged the branch I just had to pick one.

src/components/AppProvider.tsx
heypoom: @WiNloSt I completely forgot that we can pass a dynamic `jwtProviderUri`. Although, that won't work in the current SDK because of the `loginStatus.status` check in the current SDK - we only allow SDK auth config to be updated if the status is not yet initialized.

(It looks like that this check in the SDK is written incorrectly too - you aren't supposed to `return` early in `useEffect`, as returning is for sending the context)

```tsx
useEffect(() => {
  if (loginStatus.status !== "uninitialized") {
    return;
  }

  setupSdkAuth(config, dispatch);
}, [config, dispatch, loginStatus.status]);
```

WiNloSt: Let's keep this as is then. We should fix this separately.

src/routes/product-list/ProductCard.tsx
heypoom: @WiNloSt Thanks for testing this! The CSS was indeed broken, I just fixed that in https://github.com/metabase/shoppy/pull/71/commits/7ac8830987ed1c3b8d9a71c37958c5f9a04f28cc. You should be able to link the `src` to a nonexistent image or just remove the `src`, and it should show up like in this screenshot:

CleanShot 2567-10-09 at 22 57 45@2x

It's less important now that we have the preload, but still good to have in case of a slow network connection.

PR #72, opened 2024-10-09
fix: do not close the mobile menu when clicking on parent link

heypoom:
When clicking on parent links (i.e. Products and Analytics) on mobile, it should only collapse and show the links and not close the mobile menu.

### How to test

- Set the screen to mobile viewport on the `main` branch. Clicking on parent links will close the mobile menu.
- On this branch, the mobile menu stays open.

### DIFF
--- a/src/components/layout/Shell.tsx
+++ b/src/components/layout/Shell.tsx
@@ -82,7 +82,13 @@ export function Shell(props: Props) {
                 <SiteLogo />
               </Link>
 
-              <SidebarLinks onLinkClick={closeMobileNav} />
+              <SidebarLinks
+                onLinkClick={(link) => {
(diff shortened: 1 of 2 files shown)


src/components/layout/Shell.tsx
```diff
@@ -82,7 +82,13 @@ export function Shell(props: Props) {
                 <SiteLogo />
               </Link>
 
-              <SidebarLinks onLinkClick={closeMobileNav} />
+              <SidebarLinks
+                onLinkClick={(link) => {
+                  if (!link.children) {
```
heypoom, line 87: Only close the mobile menu if it's not a parent link (i.e. it doesn't have children)

PR #75, opened 2024-10-09
Hide PDF downloads on dashboards

heypoom:
PDF downloads looks broken on dashboards when custom themes are applied, especially on the line height. We should hide it from the demo for the time being.

### DIFF
--- a/src/routes/analytics/DashboardPage.tsx
+++ b/src/routes/analytics/DashboardPage.tsx
@@ -15,7 +15,11 @@ export function DashboardPage(props: Props) {
 
   return (
     <Box mih="100vh" className="dashboard-container smartscalar">
-      <InteractiveDashboard dashboardId={dashboardId} withTitle withDownloads />
+      <InteractiveDashboard
+        dashboardId={dashboardId}
+        withTitle
+        withDownloads={false}
+      />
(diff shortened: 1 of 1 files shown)


src/routes/analytics/DashboardPage.tsx
```diff
@@ -15,7 +15,11 @@ export function DashboardPage(props: Props) {
 
   return (
     <Box mih="100vh" className="dashboard-container smartscalar">
-      <InteractiveDashboard dashboardId={dashboardId} withTitle withDownloads />
+      <InteractiveDashboard
+        dashboardId={dashboardId}
+        withTitle
+        withDownloads={false}
```
npretto, line 21: This is also hiding results (json,csv etc) downloads right?
After we look into the issue tomorrow we should talk with Alberto to see if we should split the props in two.
I also expect pngs to be broken so let's also check that

heypoom: @npretto yes, that also hides the "download result" button in each card. I think splitting in two makes sense.

npretto (approved):

heypoom:
Merging this for the time being (until we fixed the PDF downloads)

Issue #53, opened 2024-07-16
Reduce stalled requests due to number of requests to the same domain


heypoom:
Shoppy requests are very slow due to a browser limitation on the number of requests to the same domain. As a result, we have these stalled requests. We tried to do exactly the same requests using Jmeter, and no issues like this. Other configurations such as caching, instance size, etc, doesn't work.

One hypothesis that we can try is to eliminate preflight requests, it should reduce the overall query time. To do this, we need to proxy all [shoppy.hosted.staging.metabase.com](http://shoppy.hosted.staging.metabase.com/) requests via https://metabase-shoppy.vercel.app/

The goal is to speed up Shoppy loading time.

PR #78, opened 2024-10-16
Load the product card question only when in viewport to prevent stalled requests

heypoom:
Closes https://github.com/metabase/shoppy/issues/53

We are loading all 20 product cards at once, even when the user might only have 1 - 3 product cards in viewport. This caused the browser to **stall the requests** by 3 - 5 seconds, as the browser has no way of knowing which questions to load first. It gets really slow when the browser loads the last question first.

This PR adds an `IntersectionObserver` to the `StaticQuestion` in the product card, which eliminates the stalled request and loads the product card gradually as we scroll.

An interesting question is whether this lazy-loading would be a nice behaviour to add to the SDK itself, or optionally via a prop in the SDK. (e.g. `<InteractiveQuestion loadWhenVisible={false} />`)

### Screenshots

Stalled Requests

CleanShot 2567-10-16 at 19 20 54@2x

Before

CleanShot 2567-10-16 at 19 20 15@2x

After

CleanShot 2567-10-16 at 19 19 09@2x

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-error-boundary": "^4.0.12",
+    "react-intersection-observer": "^9.13.1",
     "sort-by": "^1.2.0",
     "typescript": "^5.4.5",
     "vite": "^5.3.1",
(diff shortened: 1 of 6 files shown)


WiNloSt (approved):
Looks good. I don't thing we need to wrap the component inside the modal though.

Issue #79, opened 2024-10-30
Low-hanging fruit design improvements for Shoppy


heypoom:
We should make design improvements to improve the Shoppy demo, according to this [Notion product doc](https://www.notion.so/metabase/Make-low-hanging-fruit-improvements-to-Shoppy-12a69354c9018011be49edf18852c7c0?pvs=4).

PR #80, opened 2024-11-01
Design improvements for Shoppy (Part 1)

heypoom:
Closes https://github.com/metabase/shoppy/issues/79

This PR makes design improvements to Shoppy. Refer to the [product docs](https://www.notion.so/metabase/Make-low-hanging-fruit-improvements-to-Shoppy-12a69354c9018011be49edf18852c7c0?pvs=4) for the full list of changes to Shoppy - this PR acts on a subset of them, as other changes will require changes to the SDK itself. I have commented in the product doc on which change requires further improvements.

PS. the diffs is a bit misleading, the `yarn.lock` changes make it looks big but it's actually not a lot of diffs. 

### Loom

[Private Loom](https://www.loom.com/share/65cdd5b9cbd24c6da245270eea954823?sid=cc502dec-4dc2-4f1a-9194-909443b5f827)

### Checklist

- [x] Sidebar navigation copy
- [x] Remove the divider from sidebar
- [x] Add "New custom exploration" and "New dashboard" buttons
- [x] Small copy tweaks
- [x] Implement the "add a filter" and "change the summary" buttons by using modular `InteractiveQuestion` components
- [x] Re-implement the “create question from template” flow with InteractiveQuestion instead of ModifyQuestion


src/components/CustomIcon.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const CustomIcon = ({
+  icon,
+  size = 16,
+  fill,
+  ...props
+}: CustomIconProps) => (
+  <svg
+    viewBox="0 0 16 16"
```
heypoom, line 23: So far our icons here only uses the `16x16` viewBox.

src/components/InteractiveQuestionView.tsx
```diff
@@ -0,0 +1,105 @@
...
+interface Props {
+  isSaveEnabled?: boolean
+}
+
+export const InteractiveQuestionView = (props: Props) => {
+  const { isSaveEnabled = false } = props
+
+  const [view, changeView] = useReducer(
+    (view: QuestionView, nextView: QuestionView) =>
+      view !== "viz" && view === nextView ? "viz" : nextView,
```
heypoom, line 21: This allows us to click on the same view's button again and go back to the visualization view.

src/components/InteractiveQuestionView.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+// TODO: add the "breakout" view once we've separated the breakout from summarization in the SDK
+type QuestionView = "viz" | "filter" | "summary" | "editor"
+
+interface Props {
+  isSaveEnabled?: boolean
+}
+
+export const InteractiveQuestionView = (props: Props) => {
+  const { isSaveEnabled = false } = props
```
heypoom, line 17: We only enable the "Save" button in the "new question from template" page for now.

src/components/InteractiveQuestionView.tsx
```diff
@@ -0,0 +1,105 @@
...
+          )}
+
+          <ThemedButton size="compact-sm" onClick={() => changeView("editor")}>
+            <CustomIcon icon="notebook" />
+          </ThemedButton>
+        </Group>
+      </Flex>
+
+      {view === "viz" && (
+        <Box h="500px">
```
heypoom, line 73: We use a fixed height for now, as the SDK does not provide a sensible default height in `QuestionVisualization` yet. We can remove this from Shoppy once the SDK implements a sane default height in `QuestionVisualization`.

src/store/site.ts
```diff
@@ -13,3 +14,5 @@ export const siteAtom = atomWithStorage<SiteKey>(
     getOnInit: true,
   },
 )
+
+export const siteIsReloadingAtom = atom(false)
```
heypoom, line 18: We need to fully reload the page in some cases - do a full page search on `useReloadOnSiteChange` to see where that applies. This helps us to show the full-page loading indicator while the page is being reloaded by `window.location.reload()`

src/components/InteractiveQuestionView.tsx
WiNloSt: why do you destructure props on some components but not all? I can understand in case you need to access the whole `props` for spreading for example. But I don't see the reason to do that here. I found using `props` needs a bit more work to figure out where the all the props are used.

src/components/InteractiveQuestionView.tsx
```diff
@@ -0,0 +1,105 @@
...
+    <Box>
+      <Flex
+        w="100%"
+        justify="space-between"
+        pb={14}
+        direction={{ base: "column", sm: "row" }}
+        rowGap="sm"
+      >
+        <Group gap="xs">
+          <InteractiveQuestion.BackButton />
```
WiNloSt, line 38: How do you get this button to render? I don't couldn't figure out how.

src/components/InteractiveQuestionView.tsx
WiNloSt: inlining this logic makes it a little hard to understand. What this does is essentially changing the view to `viz` if we're clicking the same button as the current view. Otherwise, change to whatever the nextViz is.

I'd like to see this being structured with some if blocks, that might make this more readable.

src/components/InteractiveQuestionView.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+      {isSaveModalOpen && (
+        <Modal
+          opened={isSaveModalOpen}
+          onClose={closeSaveModal}
+          classNames={{ content: "bg-background" }}
+          withCloseButton={false}
+        >
+          <Box bg="background">
+            <InteractiveQuestion.SaveQuestionForm onClose={closeSaveModal} />
```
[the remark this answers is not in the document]
WiNloSt: Is there a way to automatically close the modal once the question is saved? mine showed `success` then changed to `save` again after a while, but the modal remains which is not expected.

WiNloSt (approved):
Tested and all looks good 👍

src/components/InteractiveQuestionView.tsx
heypoom: @WiNloSt Good idea. I initially has this explicitly written with if-blocks, but later collapsed to this. I'll revert this then, as the former version is indeed more readable.

src/components/InteractiveQuestionView.tsx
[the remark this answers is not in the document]
heypoom: @WiNloSt This smells like a bug on the SDK - the `InteractiveQuestion.SaveQuestionForm` component only exposes a single `onClose` prop, so if the modal does not close after success we need to fix it. Let me re-test it.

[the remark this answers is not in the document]
WiNloSt: I'm using `51-stable` by the way which points to the same version you have here. I did that just to test the tag.

src/components/InteractiveQuestionView.tsx
heypoom: @WiNloSt Good callout, I just realized that the back button is hidden if we do not have `onNavigateBack` specified. We don't mention this in the docs either - let's update the doc to indicate this.

src/components/InteractiveQuestionView.tsx
heypoom: Whoops, bad habit. I usually do not destructure the `props` in place. Fixed in https://github.com/metabase/shoppy/pull/80/commits/9f82f07cec66a35a295b5c6d7cece69caeafc689

PR #82, opened 2024-11-08
Remove custom font workarounds from Shoppy

heypoom:
Closes https://github.com/metabase/metabase/issues/44394

Removes the custom font workarounds we were using, as the SDK didn't support custom fonts before.


public/fonts/pug.font.css
```diff
@@ -1,16 +1,8 @@
...
-  font-style: normal;
-  font-weight: 400;
-  font-display: swap;
-  src: url(https://fonts.gstatic.com/s/dmmono/v14/aFTU7PB1QTsUX8KYthqQBK6PYK0.woff2) format('woff2');
-  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+0304, U+0308, U+0329, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
-}
-
-@font-face {
-  font-family: 'Custom';
+  font-family: 'DM Mono';
```
heypoom, line 5: We still need the font-face declaration here that aliases the **500** weight to the **700** weight, because Metabase uses the 700 weight everywhere. We should add font weight customization as a theme option in the SDK 😥

npretto (approved):

Issue #85, opened 2024-11-12
Use the InteractiveQuestion component for creating questions from scratch in Shoppy


heypoom:
Use the `<InteractiveQuestion />` sdk component from creating questions, instead of `<CreateQuestion />`.

See [this product doc section](https://www.notion.so/metabase/Make-low-hanging-fruit-improvements-to-Shoppy-12a69354c9018011be49edf18852c7c0?pvs=4#12a69354c90180b29598e899df461fcb) for more information.

PR #86, opened 2024-11-12
Use the InteractiveQuestion component to create questions

heypoom:
Closes https://github.com/metabase/shoppy/issues/85

### DIFF
--- a/src/components/CreateQuestionView.tsx
+++ b/src/components/CreateQuestionView.tsx
@@ -1,0 +1,67 @@
+import { Box, Flex, Group, Modal } from "@mantine/core"
+import { InteractiveQuestion } from "@metabase/embedding-sdk-react"
+
+import { ThemedButton } from "./ThemedButton"
+import { useDisclosure } from "@mantine/hooks"
+import { useAtom } from "jotai"
+import { isSaveModalOpenAtom } from "../store/save"
+
+export const CreateQuestionView = () => {
+  const [isVisualization, { toggle }] = useDisclosure(false)
+  const [isSaveModalOpen, setSaveModalOpen] = useAtom(isSaveModalOpenAtom)
+
+  return (
+    <Box>
+      <Flex
+        w="100%"
+        justify="space-between"
+        pb={14}
+        direction={{ base: "column", sm: "row" }}
+        rowGap="sm"
+      >
+        <Group gap="xs">
+          <InteractiveQuestion.Title />
+        </Group>
+
+        <Group gap="xs">
+          <ThemedButton size="compact-sm" onClick={toggle}>
+            Show {isVisualization ? "editor" : "visualization"}
+          </ThemedButton>
+
+          <ThemedButton
+            size="compact-sm"
+            onClick={() => setSaveModalOpen(true)}
+          >
+            Save
+          </ThemedButton>
+
+          <InteractiveQuestion.ResetButton />
+        </Group>
+      </Flex>
+
+      {isVisualization && (
+        <Box h="500px">
+          <InteractiveQuestion.FilterBar />
+          <InteractiveQuestion.QuestionVisualization />
+        </Box>
+      )}
+
+      {!isVisualization && <InteractiveQuestion.Editor onApply={toggle} />}
+
+      {isSaveModalOpen && (
+        <Modal
+          opened={isSaveModalOpen}
+          onClose={() => setSaveModalOpen(false)}
+          classNames={{ content: "bg-background" }}
+          withCloseButton={false}
+        >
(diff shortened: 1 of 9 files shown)


src/components/CreateQuestionView.tsx
```diff
@@ -0,0 +1,85 @@
...
+    useState(false)
+
+  const onApplyEditorChanges = () => {
+    toggle()
+
+    // Do not show the "show visualization/editor" button until the user has visualized the question.
+    // This prevents users from seeing the "Question not found" indicator.
+    if (!isVisualization) {
+      setVisualizationButtonShown(true)
+    }
```
heypoom, line 24: This prevents the user for accidentally seeing "Question not found", or an empty screen because visualization is not ready yet.

PS. I think we should not show "Question not found" in the SDK if `questionId` is null or undefined. This logic would be nevertheless needed to avoid a blank screen.

src/components/CreateQuestionView.tsx
```diff
@@ -0,0 +1,88 @@
...
+      {isSaveModalOpen && (
+        <Modal
+          opened={isSaveModalOpen}
+          onClose={() => setSaveModalOpen(false)}
+          classNames={{ content: "bg-background" }}
+          withCloseButton={false}
+        >
+          <Box bg="background">
+            <InteractiveQuestion.SaveQuestionForm
+              onClose={() => setSaveModalOpen(false)}
```
heypoom, line 81: - I just noticed this `onClose` prop is called only when the form is _cancelled_, not when it has succeeded.
- I also noticed that the two buttons does not look equal in size. We should fix that in the SDK as well.
- I wonder if we can have a hook that creates the question based on this question context (that can be used in this `CreateQuestionView` component for instance), so users does not have to use our `SaveQuestionForm` component.

src/components/CreateQuestionView.tsx
```diff
@@ -0,0 +1,88 @@
...
+    <Box>
+      <Flex
+        w="100%"
+        justify="space-between"
+        pb={14}
+        direction={{ base: "column", sm: "row" }}
+        rowGap="sm"
+      >
+        <Group gap="xs">
+          <InteractiveQuestion.Title />
```
heypoom, line 37: Initially shows "New Question", then changes to the name of the newly-saved question.

src/components/CreateQuestionView.tsx
```diff
@@ -0,0 +1,87 @@
+import { useAtom } from "jotai"
```
npretto, line 1: Is this how people are supposed to create questions with InteractiveQuestion?
Shouldn't we provide a "works out of the box" component and use it here? Or do we need something custom for shoppy?

heypoom: @npretto we need the custom views for the custom themed buttons ([see the product docs](https://www.notion.so/metabase/Make-low-hanging-fruit-improvements-to-Shoppy-12a69354c9018011be49edf18852c7c0?pvs=4#12a69354c901809f9e28fbce6c44b145)) for Shoppy, yes. Using the default component wouldn't work here as we need to demonstrate customization.

WiNloSt:
I'm not sure where the discussion around deprecating `CreateQuestion` is going, but if we're not going to deprecate it and keep it for the sake of simpler API, and easy to use. Should we, instead of nuking `CreateQuestion`, keep it and add a new way to create a question with `InteractiveQuestion` like in this PR to show our users how we could customize the create question flow to fit their use case?

src/components/CreateQuestionView.tsx
npretto: We need to re-implement all of it just to change the style/position on the buttons? 🤯

heypoom: @npretto With the current `InteractiveQuestion` API or even with the upcoming `CreateQuestion` API, yeah there is no way around this. We'll be implementing the custom picker components (e.g. `FilterPicker`) and more though, so in this specific case we can't get around implementing a custom view.

heypoom:
@WiNloSt I think we have a consensus ([see Slack discussion here](https://metaboat.slack.com/archives/C063Q3F1HPF/p1731683699167609?thread_ts=1731559962.795409&cid=C063Q3F1HPF)) to not deprecate the component and change `CreateQuestion` to use InteractiveQuestion directly (https://github.com/metabase/metabase/issues/50082) - we'll be updating the API, implementation and docs.

(That won't affect this PR though)

npretto (approved):

heypoom:
Getting this in first for the Metabase 51 Release Webinar

PR #88, opened 2024-11-29
Fix interactive question in modal being cut off

heypoom:
Fixes the Shoppy modal from being cut-off when clicking on the chart in the product page, and bumps the sdk version. This appears to be a modal-specific issue.

CleanShot 2567-11-30 at 02 31 35@2x

### DIFF
--- a/src/components/InteractiveQuestionView.tsx
+++ b/src/components/InteractiveQuestionView.tsx
@@ -77,7 +77,7 @@ export const InteractiveQuestionView = ({ isSaveEnabled = false }: Props) => {
       </Flex>
 
       {view === "viz" && (
-        <Box h="500px">
+        <Box>
           <InteractiveQuestion.FilterBar />
           <InteractiveQuestion.QuestionVisualization />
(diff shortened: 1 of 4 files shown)


src/components/InteractiveQuestionView.tsx
```diff
@@ -77,7 +77,7 @@ export const InteractiveQuestionView = ({ isSaveEnabled = false }: Props) => {
       </Flex>
 
       {view === "viz" && (
-        <Box h="500px">
+        <Box h={500}>
```
WiNloSt, line 80: Just wanna confirm. This seems to change nothing as the new value will basically result in the same pixel value.

WiNloSt (approved):
Tested. Now the chart modal isn't capped anymore.

src/components/InteractiveQuestionView.tsx
heypoom: yep, I'm not sure why I changed that to {500}, let me revert that

PR #91, opened 2024-12-23
Add ProficiencyLabs Shoppy theme

heypoom:
Closes https://github.com/metabase/metabase/issues/51629

Adds a new "ProficiencyLabs" Shoppy theme that emulates a typical B2B or SaaS use cases, and makes it the default theme. Should closely match [the following Figma draft design](https://www.figma.com/design/OamRqg07JOoC4bUqfCkh4t/SDK-Reference-APP?node-id=2656-4365&t=cUPVsA3UuKro8BhM-4). I have left plenty of inline review comments with screenshots as this PR is a bit big, to make it easier to review.

# Screenshots

CleanShot 2568-01-07 at 18 25 53@2x

CleanShot 2568-01-07 at 18 26 02@2x

CleanShot 2568-01-07 at 18 26 46@2x


src/components/layout/SidebarLinks.tsx
```diff
@@ -25,9 +27,33 @@ const renderLink = (
 ) => {
   const { isChild, onLinkClick } = context ?? {}
 
+  function renderIcon(): ReactNode {
```
heypoom, line 30: Sidebar icons for the Acme theme.

src/index.tsx
```diff
@@ -15,14 +15,14 @@ import "./styles/index.css"
...
 // Theme overrides for each themes
+import "./themes/acme.css"
 import "./themes/stitch.css"
 import "./themes/luminara.css"
 import "./themes/pug.css"
 
 // CSS style overrides used for workarounds.
 import "./styles/smartscalar-workaround.css"
 import "./styles/dashboard-workaround.css"
-import "./styles/viz-workaround.css"
```
heypoom, line 25: This workaround was breaking some dashcards on mobile, and it's no longer needed.

src/routes/analytics/DashboardPage.tsx
```diff
@@ -19,6 +20,11 @@ export function DashboardPage(props: Props) {
         dashboardId={dashboardId}
         withTitle
         withDownloads={false}
+        renderDrillThroughQuestion={() => (
+          <Container size="1000px" w="100%">
+            <InteractiveQuestionView isSaveEnabled={false} />
+          </Container>
+        )}
```
heypoom, line 27: Uses the themed drill-throughs instead of the default views, which looks weird on Acme theme.

src/routes/internal/KitchenSink.tsx
```diff
@@ -23,7 +23,7 @@ export const KitchenSink = () => (
     <Grid pb={20}>
       {questions.map((q) => (
         <Grid.Col key={q} span={4}>
-          <StaticQuestion questionId={q} showVisualizationSelector={false} />
+          <StaticQuestion questionId={q} withChartTypeSelector={false} />
```
heypoom, line 26: Applies renames from the new SDK version.

src/routes/product-detail/ProductDetailCard.tsx
```diff
@@ -29,34 +29,8 @@ export const ProductDetailCard = ({ product }: Props) => {
...
-
-            <Flex>
-              <Text w="100%" c="light-grey">
-                discount
-              </Text>
-              <Text>$4.00</Text>
-            </Flex>
+          <Box>
+            <Text>{product.description}</Text>
           </Box>
```
heypoom, line 34: Uses a proper product description instead of the hard-coded product insights, which does not make sense for themes other than theStitch.

src/routes/product-list/index.tsx
```diff
@@ -25,22 +26,32 @@ export const ProductAnalyticsPage = (props: Props) => {
...
+  })
+
   let products = query.data ?? []
 
   if (categoryId) {
     products = products.filter((product) => product.category.id === categoryId)
   }
 
+  const currentCategoryName =
+    categoryId && categoryQuery.data?.find((c) => c.id === categoryId)?.name
```
heypoom, line 41: Displays product categories we are currently in, rather than always showing "All products"

src/styles/dashboard-workaround.css
```diff
@@ -38,4 +38,32 @@ body[data-theme="stitch"] {
       -webkit-text-fill-color: transparent;
     }
   }
+}
+
+body[data-theme="acme"] .dashboard-container {
```
heypoom, line 43: Overrides that are needed for dashboards on Acme to match the Figma design.

WiNloSt:
General feedback
- By making this default I assume shoppy will load with the first theme, but I always get the second theme (Stitch) instead.
- One thing I noticed is that on the products page (/admin/products), if I change tab to another tab on the same browser window and back to shoppy again, it will reload all the trend charts under each product. This doesn't seem to happen when changing windows <kbd>Alt</kbd> + <kbd>Tab</kbd>, only switching tabs would affect this and this doesn't affect the product details page (/admin/products/:id)
- click actions popover doesn't have enough border contrast, or simply put, it looks like it's missing a border.
    image
- chart hovers have different font weights (product details page)
    image
    image
- Github logo links to https://github.com/metabase/shoppy, I think this is desirable, but just want to double check that we don't want to link to other URL like metabase repo.
- the book/doc icon links to https://www.metabase.com/docs/latest/embedding/start, since this is SDK focused demo, should we link to the [SDK page](https://www.metabase.com/docs/latest/embedding/sdk/introduction) instead?
- Would it be possible to generate more images, so there are no same images sitting next to each other on any page?
    image
- It's not apparent that we can toggle these headers, other themes provide visual cues for that, but not in the new theme
    image
    image
- The footer (I hope it's meant to be a footer) doesn't always stay at the bottom of the page
    Screenshot 2025-01-07 at 8 34 24 PM
- When saving a question or a dashboard, the default collection changes whether what theme is first loaded. Try to visit the first theme and refresh vs visit the other theme, refresh, and visit the first theme
    image
    image

public/logo-acme.svg
WiNloSt, line 1: This one too, should we make it inline? Some SVG are inline, and some are URLs at the moment.

WiNloSt (commented):
16/49 files review. I'll continue tomorrow.

heypoom:
(Removing reviews for now so I can address Alessio and Kelvin's feedback first, will add reviewers once the feedbacks are addressed)

AlessioLaiso:
Following up on Kelvin's points above:

- Click actions popover border: let's use the same border color that's in place for the cards.
- Chart hover font weights and styles: let's use bold all caps for the header, regular for the name and bold for the numbers.
- Footer links: see [this Figma](https://www.figma.com/design/OamRqg07JOoC4bUqfCkh4t/SDK-Reference-APP?node-id=2611-18273&t=w9JBItiSArl2nLrn-1).
- Headers toggles: these are actually **not** meant to be togglable - they should always be expanded.
- Agreed with the comments I didn't mention.
- And one more thing: let's change the "Acme Co." name to "ProficiencyLabs" as mentioned on Slack.

heypoom:
@WiNloSt @AlessioLaiso Thanks for the detailed reviews guys! Applied the changes according to your feedback here:

- GitHub logo links - yes, we link to Shoppy demo so our customers could see how it’s made.
- The book/doc icon links - updated to https://www.metabase.com/docs/latest/embedding/sdk/introduction as it makes more sense for people who are curious about SDK embedding
- Made ProficiencyLabs the default theme, and added a migration flag to switch to ProficiencyLabs users who viewed the demo before.
- Added min-height to content so the footer is at the bottom of the screen
- Changing tab to another tab on the same browser window and back to shopp will reload all the trend charts under each product -- this was caused by re-rendering on the product list component, now fixed.
- Disabled toggling sidebar headers for ProficiencyLabs
- When saving a question or a dashboard, the default collection changes whether what theme is first loaded -- this was caused by `SANDBOXED_CUSTOM_ANALYTICS_COLLECTIONS` and `SANDBOXED_USER_GENERATED_COLLECTIONS` missing the `proficiency` site. Now, we never show the “Which collection should this go in” and automatically save to those collections.
- Added click action popover borders
- Used 24px padding for left and right side of the sidebar
- Updated chart hover font weights and styles to use bold all caps for the header, regular for the name and bold for the numbers
- SVG custom icons are now all inline - migrated bar and dashboard icons to inline
- Changed Acme Co. to ProficiencyLabs

For generating more images, I will do that tomorrow.

src/routes/product-list/index.tsx
```diff
@@ -11,36 +11,47 @@ import { useAtom } from "jotai"
...
 
 interface Props {
   categoryId?: string
 }
 
 export const ProductAnalyticsPage = (props: Props) => {
   const [site] = useAtom(siteAtom)
-  const categoryId = props.categoryId && parseInt(props.categoryId, 10)
+
+  const categoryId = props.categoryId ? parseInt(props.categoryId, 10) : null
```
WiNloSt, line 23: why do we need to cast from undefined to null when it's not passed?

src/routes/product-list/index.tsx
```diff
@@ -11,36 +11,47 @@ import { useAtom } from "jotai"
...
+
   let products = query.data ?? []
 
   if (categoryId) {
     products = products.filter((product) => product.category.id === categoryId)
   }
 
-  // If the site changes, redirect back to the product listing page.
+  const currentCategoryName =
+    categoryQuery.data?.find((c) => c.id === categoryId)?.name ?? "All products"
```
WiNloSt, line 43: nit: use full name

```suggestion
    categoryQuery.data?.find((category) => category.id === categoryId)?.name ?? "All products"
```

src/utils/sidebar-links.tsx
```diff
@@ -0,0 +1,105 @@
...
+        ),
+      },
+      {
+        title: "Analytics",
+        children: [
+          {
+            to: "/admin/analytics",
+            title: "Dashboards",
+            icon: () => (
+              <CustomIcon
```
WiNloSt, line 71: it was really hard to understand why this wasn't rendered for other themes, and from scrolling a long PR I found that this is defined in the theme. I'm not sure if this is the best way to do this.

Maybe it's okay, but I was just confused why this icon isn't rendered for other themes. I wish it's structured in a way that you can put these custom icons in a theme-specific file that it'd be more obvious that these icons are only rendering on certin theme.

WiNloSt (approved):
I think it's in a merge-able state. Please see if there's any other feedback you can address as you see fit

Other feedback
- *Products* and *Analytics* headers are still showing click icon on hover, but clicking them doesn't collapse the list which is now good.
- echarts tooltip is still missing a border, it's not that apparent, but the style also looks different, with a more pronounced background shadow. I wouldn't do anything here if it's proven to be complicated, but it'd be nice if we could get the style more consistent.
    Screenshot 2025-01-09 at 3 17 39 PM
- Whether opening or closing a modal by clicking a bar in the bar chart inside the product details page would reload the bar chart on the product details page (not the one inside the modal)

src/styles/index.css
```diff
@@ -60,4 +64,13 @@ body:not([data-theme="luminara"]) {
...
+}
+
+.sidebar-links-container {
+  @apply space-y-3;
+}
+
+/** Mantine scroll lock causes the header to shift right by 15px on mobile when opening a modal. This class prevents that. */
+.preserve-header-position.preserve-header-position {
+  right: 0 !important;
+}
```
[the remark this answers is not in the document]
heypoom: @WiNloSt I couldn't reproduce this issue now even on that exact commit, so I removed this class. There used to be a class that moves the top navigation bar 15px to the right as part of a scroll locking library whenever the "create dashboard" modal was open. I have no clue why the issue is suddenly gone. Thanks for catching this!

src/routes/product-list/index.tsx
heypoom: This is just to make the type a bit nicer as the type used to include empty string in the union.

CleanShot 2568-01-09 at 18 21 25@2x

src/utils/sidebar-links.tsx
heypoom: That was indeed confusing - thanks for catching this! Fixed in https://github.com/metabase/shoppy/pull/91/commits/846b7791e0b9acfab0c4aab25c8998cdcbcb83bc by making the icons object site-specific (i.e. `{ icons: { proficiency: "foo" } }`)

src/components/layout/SidebarLinks.tsx
```diff
@@ -1,33 +1,71 @@
...
+  const { isChild, onLinkClick, site } = context ?? {}
+
+  function renderIcon(): ReactNode {
+    if (typeof link.icon === "string") {
+      return <Icon icon={link.icon} />
+    }
+
+    if (!link.icon) {
+      return null
+    }
```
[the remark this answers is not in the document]
heypoom: Good catch, fixed in https://github.com/metabase/shoppy/pull/91/commits/64883d3160b4f9d7c9a8ba7ddbd5477630e8a9bd

src/components/SiteLogo.tsx
```diff
@@ -1,11 +1,30 @@
...
+      <Box>
+        <Flex justify="flex-start" align="center" pb="20px">
+          <Image src="/logo-proficiency.svg" maw="40px" mr="8px" />
+
+          <Text fz="21px" lh="21px" fw={700}>
+            ProficiencyLabs
+          </Text>
+        </Flex>
+
+        <Divider
```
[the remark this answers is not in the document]
heypoom: Good catch, fixed in https://github.com/metabase/shoppy/pull/91/commits/c1684aba3841b08fbfd620093972405874c47e84

Issue #83, opened 2024-11-11
Use new default question layout in Shoppy once they're released


heypoom:
# Blocked

Blocked on https://github.com/metabase/metabase/issues/52261

Once the modular querying components in https://github.com/metabase/metabase/issues/49713 is released, we should update Shoppy. Once we have popover, modal, and sidebar components for filtering, summarizing, and breaking out, as shown in the Loom recording over to the right, we should use them for these filter, summarize, and breakout themed buttons:

See the [product doc](https://www.notion.so/metabase/Make-low-hanging-fruit-improvements-to-Shoppy-12a69354c9018011be49edf18852c7c0?pvs=4#12a69354c90180bd8fbffb91ee2d4bba) for more info.

PR #94, opened 2025-01-14
Use default question layout and adjust theming

heypoom:
Closes https://github.com/metabase/shoppy/issues/83


WiNloSt:
Can you help check if this is an issue with this PR, re: missing button, or something else.

Screenshot 2025-02-06 at 1 50 40 PM
Screenshot 2025-02-06 at 1 50 44 PM

src/themes/luminara.css
```diff
@@ -92,5 +92,12 @@ body[data-theme="luminara"] {
   .themed-button {
     background: var(--color-light-grey);
   }
+
+  .question-container, .dashboard-container {
+    /* WORKAROUND: table border radius */
+    [data-testid="TableInteractive-root"] {
+      border-radius: 6px;
+    }
```
WiNloSt, line 100: Is this more like a problem we should fix at the SDK level instead of patching it here?

Do we have a task for this yet?

WiNloSt (approved):
Tested every component it looks great, I found one invisible button when not hovering tho. It in the table conditional formatting or something.

src/themes/luminara.css
heypoom: @WiNloSt This is applying styles so they match the Luminara theme's styles of rounded corners.

I think we have to consider if we want to expose a CSS class instead of a theme option too.

WiNloSt: so our tables are always without rounded corner?

heypoom: Yep, current styles is no border radius afaik

PR #97, opened 2025-02-06
Fix cutoff on shell layout

oisincoveney:
Fixes the cutoff on the shell layout. This is most obvious when trying to create a new question, and was brought up by an SDK user.

Before:

image

After:
image

Notice the header and save button appearing.

**Note** when testing, play around with the whole app and make sure that the padding between the navbar and the main content seems reasonable. I've removed the the `pt` values from most of the components and replaced it with some padding on the main app


heypoom (approved):

heypoom:
Looks great from my testing in deploy previews - thanks Oisin!

PR #98, opened 2025-02-11
Read port from env variable

sanex3339:
Read port from env variable.
I need it for e2e testing to pass a different port


npretto (approved):

vite.config.ts
```diff
@@ -19,4 +20,7 @@ export default defineConfig({
     open: false,
     port: 3004,
   },
+  preview: {
+    ...(port && { port }),
```
sanex3339, line 24: If not set - the default (4173) value is used

sanex3339: actually let's set it's explicitly

PR #115, opened 2025-04-21
Adjust Shoppy migrations/seeding scripts for the new deployment approach.

sanex3339:
Adjust Shoppy migrations/seeding scripts for the new deployment approach.

How to verify:
- e2e CI should pass


.env.example
```diff
@@ -1,4 +1,4 @@
 PORT=4173
-VITE_APP_METABASE_INSTANCE_URL=https://shoppy.hosted.staging.metabase.com
+VITE_APP_METABASE_INSTANCE_URL=https://shoppy.coredev.metabase.com
```
sanex3339, line 2: When it is merged we still want to point the production app on Vercel to the old instance and switch to the new instance a bit later.

From what I see, we can do it via env vars on the Vercel

api/drizzle/0002_lethal_dakota_north.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE "people" ALTER COLUMN "shop_id" SET NOT NULL;--> statement-breakpoint
+DO $$ BEGIN
+ ALTER TABLE "people" ADD CONSTRAINT "people_shop_id_shops_id_fk" FOREIGN KEY ("shop_id") REFERENCES "shops"("id") ON DELETE no action ON UPDATE no action;
+EXCEPTION
+ WHEN duplicate_object THEN null;
+END $$;
```
sanex3339, line 6: Because we don't run migrations on Vercel automatically, we are ok with this change.

api/src/schema/orders.ts
```diff
@@ -1,16 +1,10 @@
...
-  timestamp,
-} from "drizzle-orm/pg-core"
+import { bigint, numeric, pgTable, timestamp } from "drizzle-orm/pg-core"
 
 import { people } from "./people"
 import { products } from "./products"
 
 export const orders = pgTable("orders", {
-  id: serial("id").primaryKey().notNull(),
+  id: bigint("id", { mode: "number" }).primaryKey().notNull(),
```
heypoom, line 7: nice, I guess this is for consistency with other models which uses `bigint` 👍🏻

api/drizzle/seeder/assign-shops.ts
```diff
@@ -1,39 +1,51 @@
...
-      columns: { id: true, productId: true },
-      where: eq(orderSchema.customerId, customer.id),
+  const mappings = await db
+    .selectDistinct({
+      customerId: orders.customerId,
+      shopId: products.shopId,
     })
+    .from(orders)
+    .innerJoin(products, eq(orders.productId, products.id))
+    .where(and(isNotNull(orders.productId), isNotNull(products.shopId)))
```
heypoom, line 19: awesome! thanks for batching this

heypoom (approved):
🎉

api/src/schema/orders.ts
sanex3339: If IRC on the supabase DB we have `bigint`

PR #118, opened 2025-04-22
Adjust e2e to run with app db dump from s3

sanex3339:
Adjust e2e to run with app db dump from s3


.github/workflows/e2e-tests.yml
```diff
@@ -4,10 +4,14 @@ on:
...
     branches:
       - "main"
+      - "release"
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
     branches:
       - "main"
+      - "release"
+      # For testing
+      - '**'
```
sanex3339, line 14: Todo: remove it before merging

oreonl (approved):

.github/workflows/e2e-tests.yml
```diff
@@ -16,10 +18,26 @@ jobs:
     name: e2e-tests
     env:
       PREMIUM_EMBEDDING_TOKEN: ${{ secrets.ENTERPRISE_TOKEN }}
+      METABASE_APP_DB_USER: ${{ secrets.APP_DB_USER }}
+      METABASE_APP_DB: ${{ secrets.APP_DB_SHOPPY_DATABASE }}
+    permissions:
+      id-token: write
```
heypoom, line 24: I haven't seen this before, but it should be safe.

> The id-token: write permission allows the GITHUB_TOKEN to make an API call to generate an OIDC token (which is an extra security token, a standard one). The OIDC token is not dangerous in itself. It is minted and signed by GitHub and includes "claims"

heypoom (approved):

.github/workflows/e2e-tests.yml
sanex3339: it's the same as with the `deployment` PR

PR #121, opened 2025-04-24
Update SDK to 0.54.8 and use entity ids instead of numeric ids

sanex3339:
Update SDK to 0.54.8 and use entity ids instead of numeric ids

After https://github.com/metabase/metabase/pull/57073 we can use entity ids without flickering


heypoom (approved):
approved the re-revert - just make sure Oisin's PR is backported and released to NPM first, then we must manually bump the SDK version in this PR to that version including the entity id flicker fixes.

PR #131, opened 2025-04-28
Deploy to Vercel via GH workflows

sanex3339:
Deploy to Vercel via GH workflows

### DIFF
--- a/.github/workflows/deploy-to-vercel.yml
+++ b/.github/workflows/deploy-to-vercel.yml
@@ -1,0 +1,58 @@
+name: Vercel Deployment
+
+on:
+  workflow_call:
+    inputs:
+      vercel-environment:
+        required: true
+        type: string
+    secrets:
+      vercel-token:
+        required: true
+      vercel-org-id:
+        required: true
+      vercel-project-id:
+        required: true
+      github-token:
+        required: true
+
+jobs:
+  deploy:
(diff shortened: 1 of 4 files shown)


vercel.json
```diff
@@ -1,5 +1,11 @@
 {
   "version": 2,
+  "git": {
+    "deploymentEnabled": {
+      "*": false,
+      "shoppy-metabot": true
```
sanex3339, line 6: We still want to auto-deploy `shoppy-metabot` branch

.github/workflows/deploy-to-vercel.yml
```diff
@@ -0,0 +1,58 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Vercel CLI
+        run: npm install --global vercel@latest
+        shell: bash
+
+      - name: Pull Vercel Environment Information
+        run: vercel pull --yes --environment=${{ inputs.vercel-environment }} --token=${{ secrets.vercel-token }}
```
heypoom, line 30: nit: I wonder if we could re-export the `secrets.vercel-token` once to an env so we don't need to pass `--token=` to every command. I'm not sure if Vercel's CLI takes a token env though, but I feel it should

heypoom (approved):

PR #132, opened 2025-04-30
Fix sandboxing for e2e and add tests

sanex3339:
Fix sandboxing for e2e and add tests
- also PR restores the `import` data fallback to be able for customers to launch the app in Docker.

### DIFF
--- a/e2e/test/compatibility.cy.spec.js
+++ b/e2e/test/compatibility.cy.spec.js
@@ -116,7 +116,7 @@ describe("Embedding SDK: shoppy compatibility", () => {
     })
   })
 
-  it("should open user created dashboards", () => {
+  it("should open user created collection", () => {
     cy.visit({
       url: "/admin",
     })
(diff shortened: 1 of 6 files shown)


metabase/entrypoint.sh
```diff
@@ -32,4 +37,18 @@ done
 
 echo "Metabase is healthy. Running additional configuration..."
 
+if [ ! -f $DUMP ]; then
```
sanex3339, line 40: Restored it back for customers to be able to run it with their Embedding token

metabase/entrypoint.sh
```diff
@@ -3,24 +3,29 @@ set -e
...
-    ;;
-  *)
-    echo "Unknown dump format..."
-    ;;
-esac
+if [ -f $DUMP ]; then
+  echo "Restoring MB App DB from dump...";
+
+  # reset `MB_CONFIG_FILE_PATH` value to prevent initialization from config
+  export MB_CONFIG_FILE_PATH=''
```
sanex3339, line 10: Fixes broken sandboxing during e2e

e2e/test/compatibility.cy.spec.js
```diff
@@ -116,17 +116,53 @@ describe("Embedding SDK: shoppy compatibility", () => {
...
+    })
 
-      expect(cy.findByText("This collection is empty").should("exist"))
+    cy.then(() => {
+      const unique = new Set(counts)
+
+      expect(
+        unique.size,
+        `All counts: [${counts.join(", ")}]`,
+      ).to.be.greaterThan(1)
```
sanex3339, line 165: In theory, random seeding can generate an equal number of orders for all four shops (405 for each). 

Should we care about it? I assume the chance is quite low.

As an easy solution, I can set the number of seeded shops to `(4 * some value) - 1` that guarantees that at least one shop will have a different number of orders.

heypoom: > As an easy solution, I can set the number of seeded shops to (4 * some value) - 1 that guarantees that at least one shop will have a different number of orders.

That sounds good! If it's not tricky let's do it

e2e/test/compatibility.cy.spec.js
```diff
@@ -116,17 +116,53 @@ describe("Embedding SDK: shoppy compatibility", () => {
     })
   })
 
-  it("should open user created dashboards", () => {
+  it("should open user created collection", () => {
```
heypoom, line 119:
```suggestion
  it("should contain the user-generated collection in saved explorations", () => {
```

since we're not actually opening the User-Generated collection

heypoom (approved):

PR #137, opened 2025-05-13
Fix seeding data types and adjust docker commands

sanex3339:
Fix seeding data types and adjust docker commands

> People
> - Birthdate should be date
> - Zip should be text
> - Created at should timestamp or datetime

The `Lat long are both text sbut should be both stored as postgis points (does MB handle that well? guess so?)` is still under the question.

Also adjusted docker build process to handle `WATCH` ENV flag, so now in Docker app can be built in the `production` mode

### DIFF
--- a/.env.docker.example
+++ b/.env.docker.example
@@ -19,3 +19,5 @@ SHOPPY_DB_PORT="5432"
 SHOPPY_DB="shoppy"
 SHOPPY_DB_USER="shoppy"
 SHOPPY_DB_PASSWORD="foobarbaz"
+
+WATCH="false"
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ RUN if [ "$WATCH" != "true" ]; then \
       echo "WATCH env is not set; running production yarn build..."; \
       yarn build; \
     else \
-      echo "WATCH env iss is set; running in development mode..."; \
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -15,4 +29,11 @@ RUN if [ -d "./local-dist/embedding-sdk" ]; then \
       echo "Local embedding-sdk dist is not found in ./local-dist/embedding-sdk, skipping copy"; \
     fi
 
-CMD ["yarn", "dev", "--host"]
+RUN if [ "$WATCH" != "true" ]; then \
+      echo "WATCH env is not set; running production yarn build..."; \
+      yarn build; \
+    else \
+      echo "WATCH env iss is set; running in development mode..."; \
```
heypoom, line 36:
```suggestion
      echo "WATCH env is set; running in development mode..."; \
```

.env.docker.example
```diff
@@ -19,3 +19,5 @@ SHOPPY_DB_PORT="5432"
 SHOPPY_DB="shoppy"
 SHOPPY_DB_USER="shoppy"
 SHOPPY_DB_PASSWORD="foobarbaz"
+
+WATCH="false"
```
heypoom, line 23: Nice! Might be good to update README.md that this exists

heypoom (approved):

PR #146, opened 2025-06-19
Add Metabot compoent to main Shoppy instance

WiNloSt:
closes https://linear.app/metabase/issue/EMB-525/add-metabot-route-to-the-main-shoppy-instance

### How to verify
<details>
<summary>Old instruction, just use the <a href="https://shoppy-guafa58yc-embedding-c312e713.vercel.app/">deploy preview link</a> to test 🤦 , or expand if you wanna see the previous instruction</summary>

Connect to the main shoppy instance
1. go to `/api` folder, edit `.env`
1. set `DB_URL=postgres://<user>:<password>@<host>:5432/postgres`. Credential can be grabbed from [1password here.](https://start.1password.com/open/i?a=AEO4BXOSHNFCJLK26XA7AD522E&v=qzcmvvcryrgijigzl2f6xlabhu&i=ruynkrhiqqdm3cdlyqs4jkwvr4&h=metabaseinc.1password.com)
1. set `METABASE_INSTANCE_URL=https://shoppy.coredev.metabase.com`
1. set `METABASE_JWT_SHARED_SECRET=<secret>`. Get the secret from the instance https://shoppy.coredev.metabase.com/admin/settings/authentication/jwt, login can be found on 1password, it should be autofilled.
1. run the server `yarn dev`
1. go to back to the root folder (`/`), edit `.env`
1. set `VITE_APP_METABASE_INSTANCE_URL=https://shoppy.coredev.metabase.com`
1. set `VITE_APP_API_HOST=http://localhost:3003/api`
1. Run the app `yarn dev`
1. Now go to http://localhost:3004.
1. Navigate to "New custom exploration" > "From Scratch", you should not see Metabot
1. Set `?metabot=true` in the URL, you should see Metabot
1. Try navigating to different pages, the search parameter should remain the same, refreshing the page should also retain the search param, so when you revisit "From Scratch", you should still see Metabot unless you manually modify the search param.

</details>

### DIFF
--- a/src/components/LinkWithSearchParams.tsx
+++ b/src/components/LinkWithSearchParams.tsx
@@ -1,9 +1,23 @@
-import { Link, useSearchParams, type LinkProps } from "wouter"
+import { Link, useSearch, type LinkProps } from "wouter"
 
-export function LinkWithSearchParams({ to, href, ...props }: LinkProps) {
-  const [searchParams] = useSearchParams()
-  const hasSearchParams = searchParams.toString().length > 0
-  const newHref = `${href ?? to}${hasSearchParams ? "?" + searchParams.toString() : ""}`
+interface ForbiddenProperties {
+  /**
+   * @deprecated Please use `href` instead.
+   */
+  to?: never
+}
+
+type RemoveFields<Type, Fields extends keyof Type> = {
+  [Property in keyof Type as Exclude<Property, Fields>]: Type[Property]
+}
+
+export function LinkWithSearchParams({
+  href,
+  ...props
+}: RemoveFields<LinkProps, "to"> & ForbiddenProperties) {
+  props
+  const search = useSearch()
+  const newHref = `${href}${search ? `?${search}` : ""}`
 
   return <Link href={newHref} {...props} />
 }
--- a/src/components/LinkWithSearchParams.tsx
+++ b/src/components/LinkWithSearchParams.tsx
@@ -1,8 +1,9 @@
 import { Link, useSearchParams, type LinkProps } from "wouter"
 
-export function LinkWithSearchParams({ to, ...props }: LinkProps) {
+export function LinkWithSearchParams({ to, href, ...props }: LinkProps) {
   const [searchParams] = useSearchParams()
-  const href = `${to}?${searchParams.toString()}`
+  const hasSearchParams = searchParams.toString().length > 0
+  const newHref = `${href ?? to}${hasSearchParams ? "?" + searchParams.toString() : ""}`
 
-  return <Link href={href} {...props} />
+  return <Link href={newHref} {...props} />
 }
--- a/src/components/LinkWithSearchParams.tsx
+++ b/src/components/LinkWithSearchParams.tsx
@@ -1,0 +1,8 @@
+import { Link, useSearchParams, type LinkProps } from "wouter"
+
+export function LinkWithSearchParams({ to, ...props }: LinkProps) {
+  const [searchParams] = useSearchParams()
+  const href = `${to}?${searchParams.toString()}`
+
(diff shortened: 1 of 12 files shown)


src/components/LinkWithSearchParams.tsx
```diff
@@ -0,0 +1,9 @@
+import { Link, useSearchParams, type LinkProps } from "wouter"
+
+export function LinkWithSearchParams({ to, href, ...props }: LinkProps) {
```
WiNloSt, line 3: I've considered and tried using the hook approach by listening to `useLocation` and `useSearchParams` and programmatically set the param back. This wasn't great because the links are still rendered without the parameter, so when users open such links in new tabs, they won't have the search parameters. This fixes the problem by rendering the correct link in the first place.

src/routes/analytics/new/NewFromScratchPage.tsx
```diff
@@ -1,19 +1,29 @@
...
   return (
     <Container w="100%">
-      <InteractiveQuestion
-        questionId="new"
-        onSave={onSaveQuestion}
-        targetCollection={collectionId}
-        isSaveEnabled
-      />
+      {shouldShowMetabot ? (
+        <MetabotQuestion />
```
WiNloSt, line 18: This is the same markup as the metabot branch
https://github.com/metabase/shoppy/blob/434e9388d30185389730dbac787a985b068d6dd2/src/routes/analytics/new/NewFromScratchPage.tsx#L6-L8

src/components/LinkWithSearchParams.tsx
```diff
@@ -0,0 +1,9 @@
...
+  const [searchParams] = useSearchParams()
+  const hasSearchParams = searchParams.toString().length > 0
+  const newHref = `${href ?? to}${hasSearchParams ? "?" + searchParams.toString() : ""}`
```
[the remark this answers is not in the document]
WiNloSt: omg, there's actually a `useSearch` thanks! I'm also not familiar with `wouter` it's what already used in the project 😅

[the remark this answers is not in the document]
WiNloSt: btw `to` is an alias to `href`, it's a weird API, and it seems we use both, I think I'll clean it up and stick to only one.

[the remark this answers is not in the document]
WiNloSt: Fixed in https://github.com/metabase/shoppy/pull/146/commits/4e6abb3af63c39c7d4a5e881fcdc8d343cabb62e

I still need to check if `search` has a value, then prepend `?` since `useaSearch` doesn't take care of `?` automatically.

src/components/LinkWithSearchParams.tsx
```diff
@@ -0,0 +1,23 @@
+import { Link, useSearch, type LinkProps } from "wouter"
+
+interface ForbiddenProperties {
+  /**
+   * @deprecated Please use `href` instead.
+   */
+  to?: never
+}
+
+type RemoveFields<Type, Fields extends keyof Type> = {
```
WiNloSt, line 10: Omit doesn't play well. I got the solution [from here.](https://github.com/microsoft/TypeScript/issues/31501#issuecomment-1079728677)

src/components/layout/Shell.tsx
```diff
@@ -110,9 +110,9 @@ export function Shell(props: Props) {
           className="py-4 md:py-0"
         >
           <Box>
-            <Link to="/admin/products">
+            <LinkWithSearchParams href="/admin/products">
```
WiNloSt, line 113: for wouter`s `Link` `href` and `to` are the same thing. One is an alias for another.

PR #153, opened 2025-08-21
Adjust deploy-production.yml to use image repo and name

sanex3339:
Preparation for Shoppy deployment with Hosted Bundle

context: https://metaboat.slack.com/archives/C094TC6TP7H/p1755518212588239

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -2,18 +2,12 @@ name: Production Deployment
 
 on:
   workflow_dispatch:
-    inputs:
-      version:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-production.yml
```diff
@@ -2,19 +2,13 @@ name: Production Deployment
 
 on:
   workflow_dispatch:
-    inputs:
-      version:
-        description: 'Metabase version to deploy'
```
sanex3339, line 7: It was not used by us, because later if we deploy after merge to master a version is deployed based on a GH variable name

PR #156, opened 2025-09-02
Add the `metabot` page

sanex3339:
Add the `metabot` page

image


metabase/metabase_data.tar.gz
sanex3339, line 1: Update it to be actual, cc @heypoom

PR #195, opened 2026-07-07
Force reindex before compatibility tests run

WiNloSt:
EMB-2059: https://linear.app/metabase/issue/EMB-2059

## Summary

Metabase's appdb search-index reindex has a race that can strand a fully-built index as "pending" instead of activating it, leaving Models (e.g. "Orders", "Products") unsearchable after a fresh DB restore. This intermittently breaks `compatibility.cy.spec.js`'s "should not display tables in the data picker" test — the data picker shows an empty state because search hasn't actually finished indexing yet.

This is just a bandage — a proper fix can't be made currently.

## Root cause

This happens on a specific combination — a Metabase jar plus a restored data dump. After metabase/metabase#76551, an unrelated startup task (loading built-in Usage Analytics content) started blocking for several seconds during boot instead of running in the background. That delay causes two search-index build tasks to run at nearly the same time instead of one after another, and the index effectively gets stuck as a result — one copy ends up fully built but never marked active, the other gets marked active empty.

### How to verify

Shoppy's own CI (`MB_RUN_MODE: dev`) doesn't reliably reproduce this — the race only shows up reliably when Metabase's own e2e orchestration runs it (`MB_RUN_MODE: e2e`, the same setup used by `e2e-shoppy-tests` in the metabase repo's CI). Verify from a `metabase/metabase` checkout instead:

1. Build an EE jar:
   ```bash
   MB_EDITION=ee ./bin/build.sh
   ```
2. Build the embedding SDK package (the sample-app flow copies this in, and will fail with `ENOENT ... resources/embedding-sdk` without it):
   ```bash
   bun run build-embedding-sdk-package
   ```
3. Keep this running in a separate terminal — it builds the actual SDK bundle content the package loads at runtime:
   ```bash
   MB_EDITION=ee bun run build-hot
   ```
4. Get the Shoppy Postgres app-db dump (one-time, real coredev instance) — enable Tailscale, log in with your work email, credentials in 1Password under "Shoppy Coredev Appdb":
   ```bash
   pg_dump "postgres://{username}:{password}@{host}:{port}/{database}" > ./e2e/tmp/db_dumps/shoppy_metabase_app_db_dump.sql
   ```
5. Run the suite pointed at this branch, in GUI mode:
   ```bash
   SAMPLE_APP_BRANCH_NAME=emb-2059-force-reindex-before-tests \
   MB_EDITION=ee MB_RUN_MODE=e2e SDK_TEST_SUITE=shoppy-e2e SAMPLE_APP_ENVIRONMENT=production \
   ENTERPRISE_TOKEN=<token> METASTORE_DEV_SERVER_URL=https://token-check.staging.metabase.com \
   CYPRESS_GUI=true \
   bun run test-cypress-host-sample-apps
   ```
   `SAMPLE_APP_BRANCH_NAME` overrides the `main`-branch default the runner otherwise fetches Shoppy from. `METASTORE_DEV_SERVER_URL` matters here — without it, token/license validation silently fails against the wrong endpoint and every premium feature (including JWT SSO) reads as disabled, which looks like an unrelated "SSO has not been enabled" error. Full local-setup details: `enterprise/frontend/src/embedding-sdk-package/dev.md` in the metabase repo ("Sample Apps compatibility with Embedding SDK tests").
6. Compare against `main` (no fix) to confirm the bug is real — same command as step 5, but:
   ```bash
   SAMPLE_APP_BRANCH_NAME=main \
   MB_EDITION=ee MB_RUN_MODE=e2e SDK_TEST_SUITE=shoppy-e2e SAMPLE_APP_ENVIRONMENT=production \
   ENTERPRISE_TOKEN=<token> METASTORE_DEV_SERVER_URL=https://token-check.staging.metabase.com \
   CYPRESS_GUI=true \
   bun run test-cypress-host-sample-apps
   ```
   This should intermittently fail "should not display tables in the data picker" on a fresh restore, while step 5 (this branch) passes.

### DIFF
--- a/e2e/test/compatibility.cy.spec.js
+++ b/e2e/test/compatibility.cy.spec.js
@@ -173,7 +173,8 @@ describe("Embedding SDK: shoppy compatibility", () => {
 
   // TODO (Kelvin 2026-07-07) bandage, not a fix. Metabase's appdb search-index reindex has a
   // race that can strand a fully-built index as "pending" instead of activating it. Couldn't
-  // land the real backend fix yet (EMB-2059); this forces a reindex and waits for it first.
+  // land the real backend fix yet; this forces a reindex and waits for it first. The actual
+  // backend bug (not EMB-2059, which is just this bandage) hasn't been filed as its own ticket yet.
   // Scoped to just this test, not the whole suite — it's the only one that needs it.
   describe("data picker", () => {
     before(() => {
--- a/e2e/test/compatibility.cy.spec.js
+++ b/e2e/test/compatibility.cy.spec.js
@@ -1,3 +1,5 @@
+import { METABASE_URL, signInAsAdmin } from "../support/helpers/sign-in"
+
 const TIMEOUT = 20000
 
 describe("Embedding SDK: shoppy compatibility", () => {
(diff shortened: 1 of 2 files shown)


e2e/test/compatibility.cy.spec.js
```diff
@@ -169,30 +171,80 @@ describe("Embedding SDK: shoppy compatibility", () => {
...
-  it("should not display tables in the data picker", () => {
-    cy.visit({
-      url: "/admin/analytics/new/from-scratch",
-    })
+  // TODO (Kelvin 2026-07-07) bandage, not a fix. Metabase's appdb search-index reindex has a
+  // race that can strand a fully-built index as "pending" instead of activating it. Couldn't
+  // land the real backend fix yet (EMB-2059); this forces a reindex and waits for it first.
+  // Scoped to just this test, not the whole suite — it's the only one that needs it.
+  describe("data picker", () => {
+    before(() => {
```
WiNloSt, line 180: This test only is wrapped within a `describe` block then we add this `before` block to always call `/api/search/force-reindex`. To ensure the search index is populated before running the test.

It's a temporary hack as detailed in the comment.

e2e/test/compatibility.cy.spec.js
```diff
@@ -169,30 +171,80 @@ describe("Embedding SDK: shoppy compatibility", () => {
     })
   })
 
-  it("should not display tables in the data picker", () => {
-    cy.visit({
-      url: "/admin/analytics/new/from-scratch",
-    })
+  // TODO (Kelvin 2026-07-07) bandage, not a fix. Metabase's appdb search-index reindex has a
+  // race that can strand a fully-built index as "pending" instead of activating it. Couldn't
+  // land the real backend fix yet (EMB-2059); this forces a reindex and waits for it first.
```
npretto: > Couldn't land the real backend fix yet (EMB-2059)

Does that mean that we'll use that ticket for the real backend fix?
I think it's currently linked to this PR so it's going to get closed and people looking at this comment might think it's already been done

WiNloSt: yeah claude made their things up. The EMB-2059 is definitely our fix, and the proper BE bug hasn't been filed.

WiNloSt: Good catch, fixed it in https://github.com/metabase/shoppy/pull/195/changes/04c69075228b68f640822cec26df1cd9c2da6b22 🙏

lorem--ipsum (approved):

npretto:
I'll merge this to make CI green, Kelvin will update the comment with the correct linear issue later

WiNloSt:
> I'll merge this to make CI green, Kelvin will update the comment with the correct linear issue later

addressed in https://github.com/metabase/shoppy/pull/196

